PR #1, opened 2015-12-01
Added query options and readme

jonasfj:
@jhford, please give this a review...

@petemoore, this affects golang client, well in the future (it's fully backwards compatible). So if you want to have a look feel free, now is the time to change anything in what query string do.

--------
Basically, this adds support for defining query-string parameters. These query-string parameters are always optional, but you can't send a query-string parameter that isn't supported.

I really don't see any use-case for query-string for `continuationToken` when paging through a long list of items (tasks, index entries, etc). Notice that we can't send the `continuationToken` as payload on a get request. As this is the only use I have for now, I went for a very simple implementation.

So query-string parameters here is really just options... Think of it as an optional payload.

Generally, we'll use payload for everything, but in cases where you want to query something, maybe query string is useful. We could also specify querystring using JSON schema, but it just seems overkill, anyways, I think we can add such support later if we really want that power.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -138,6 +138,54 @@ var schema = function(validator, options) {
   };
 };
 
+
+/**
+ * Validate query-string against query.
+ *
+ * options:
+ * {
+ *   param1: /.../,               // Reg-exp pattern
+ *   param2(val) { return "..." } // Function, returns message if invalid
+ * }
+ *
+ * Query-string options not specified in options will not be allowed. But it's
+ * optional if a request carries any query-string parameters at all.
+ */
+var queryValidator = function(options = {}) {
+  return function(req, res, next) {
+    let errors = [];
+    _.forEach(req.query || {}, (value, key) => {
+      let pattern = options[key];
+      if (!pattern) {
+        // Allow the bewit key, it's used in signed strings
+        if (key === 'bewit') {
+          return;
+        }
+        // Unsupported option
+        errors.push('Query-string parameter: ' + key + ' is not supported!');
+      }
+      if (pattern instanceof RegExp) {
+        if (!pattern.test(value)) {
+          errors.push('Query-string parameter: ' + key + '="' + value +
+                      '" does not match expression: ' + pattern.toString());
+        }
+      } else {
+        let msg = pattern(value);
+        if (typeof(msg) === 'string') {
+          errors.push('Query-string parameter: ' + key + '="' + value +
+                      '" is not valid, error: ' + msg);
+        }
+      }
+    });
+    if (errors.length > 0) {
+      return res.status(400).json({
+        message: errors.join('\n'),
+      });
+    }
+    return next();
+  };
+};
+
 /**
  * Abstraction of a client with some helper methods
  *
@@ -1254,8 +1302,8 @@ var API = function(options) {
   });
   this._options = _.defaults({}, options, {
     schemaPrefix:   '',
-    paramPatterns:  {},
-    context:        []
+    params:         {},
+    context:        [],
   });
   this._entries = [];
 };
@@ -1310,8 +1358,15 @@ var STABILITY_LEVELS = _.values(stability);
  *   route:    '/object/:id/action/:param',      // URL pattern with parameters
  *   params: {                                   // Patterns for URL params
  *     param: /.../,                             // Reg-exp pattern
- *     id(val) { return "..." }                  // Function, returns message if invalid
+ *     id(val) { return "..." }                  // Function, returns message
+ *                                               // if value is invalid
  *     // The `params` option from new API(), will be used as fall-back
+ *   },
+ *   query: {                                    // Query-string parameters
+ *     offset: /.../,                            // Reg-exp pattern
+ *     limit(n) { return "..." }                 // Function, returns message
+ *                                               // if value is invalid
+ *     // Query-string options are always optional (at-least in this iteration)
  *   },
  *   name:     'identifierForLibraries',         // identifier for client libraries
  *   stability: base.API.stability.experimental, // API stability level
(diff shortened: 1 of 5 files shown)


jonasfj:
@jhford, I suspect the travis issues with node 5 is related to peerdependencies not being automatically installed in npm 3. But we should look into it...

Note: I had huge issues with `typeof` throughout the code, when printing in particular. I'm tempted to downgrade babel...  Bug in case you're interested: https://phabricator.babeljs.io/T6644

petemoore:
Thanks for doing this Jonas. Some questions!

1. Which API calls will use the query string? Are there already pull requests for the services that will use it, or will this be done later?
2. Do we need to update the JS that renders the API docs on docs.taskcluster.net to include optional query string details?
3. I see queryString parameters can be defined by name, but I don't see any mechanism to restrict their value (e.g. regex pattern, type, min/max etc). Rather than using an array of strings for declaring allowed query parameters, would it be better to use a json object where the property names in the object are the query string property names, and the values in the json object are json subschemas that define the scopes of the query parameter values? At the moment I think we can only declare the names of the allowed query string parameters, but we can't enforce anything about the allowed values.

petemoore:
OK, maybe that is overkill. You'd also need to place restrictions that the json subschemas would not allow object or array, only number or string. And maybe there is little benefit in this.

*However* maybe it would make sense to have an array of objects, with properties "name" and "description" so at least a description can be applied to each queryString parameter, and the API docs could (beautifully) render the descriptions.

src/api.js
```diff
@@ -138,6 +138,54 @@ var schema = function(validator, options) {
...
+    });
+    if (errors.length > 0) {
+      return res.status(400).json({
+        message: errors.join('\n'),
+      });
+    }
+    return next();
+  };
+};
+
```
jhford: I like validation being built in from the start!  This seems like a good approach to me, I like it.

jhford: Also liking the focus on good error messages, thanks!

src/api.js
```diff
@@ -138,6 +138,54 @@ var schema = function(validator, options) {
...
+ * options:
+ * {
+ *   param1: /.../,               // Reg-exp pattern
+ *   param2(val) { return "..." } // Function, returns message if invalid
+ * }
+ *
+ * Query-string options not specified in options will not be allowed. But it's
+ * optional if a request carries any query-string parameters at all.
+ */
+var queryValidator = function(options = {}) {
```
jhford: Nit: if we're using var x = function () {} to ensure that a function is declared before use, we're not getting that guarantee since the var x is hoisted.  When it's hoisted, we actually get a weird situation that when the function is tried to be used, it is undefined and an appropriate error message is generated.

Let's stick to either let x = function() {} or function x() {}

test/api/route_test.js
```diff
@@ -143,6 +156,40 @@ suite("api/route", function() {
...
+    var url = 'http://localhost:23525/query-param/';
+    return request
+      .get(url)
+      .query({nextPage: 'abc'})
+      .end()
+      .then(function(res) {
+        assert(!res.ok, "Expected request failure!");
+        assert(res.status === 400, "Expected a 400 error");
+      });
+  });
```
jhford: Can we add a test where we actually do something with a url value of "http://localhost:$port/endpoint?querykey=queryvalue and ensure that the api actually gets that?  Maybe that's overkill, but it'd be nice to have.

jhford:
I like this and the approach.  I also found the typeof stuff kind of obnoxious... i'll trust your judgement on that.  I wish npm had an option like "npm install --save-with-peer-deps <pkg>" that would install the package and use the same semver specifier that pkg has for its peer deps.

jonasfj:
Thanks for doing this Jonas. Some questions!

> Which API calls will use the query string? Are there already pull requests for the services that will use it, or will this be done later?

Listing things in index api... List tasks by taskgroupId, both of these will have continuation token, that one will have to provide to get page two... Ie. paging.

> Do we need to update the JS that renders the API docs on docs.taskcluster.net to include optional query string details?

Yes, this should be easy though.

> I see queryString parameters can be defined by name, but I don't see any mechanism to restrict their value (e.g. regex pattern, type, min/max etc).

The restriction is server side, so the regexp is not defined in the JSON reference.

> However maybe it would make sense to have an array of objects, with properties "name" and "description" so at least a description can be applied to each queryString parameter, and the API docs could (beautifully) render the queryString parameter descriptions.

I'm tempted... but will we really need it... I suspect we'll rarely use queryString options... So documenting them in text makes more sense... If we find that queryString parameters are used often, then we can add this... And if we use querystring often we should probably just replace the whole thing and say that a JSON object is rendered to querystring, so use JSON schema to describe the querystring options. 

This is lightweight support, we can extend it to support required parameters etc. in the future. I think that should be possible.

@petemoore, This lightweight support is not without downsides, can we live with those. I'm willing to as the set of use-cases is small (that's why I don't bring out the heavy tools). But I want to make sure you are okay with this.

src/api.js
jonasfj: Agree `let` is always better... thanks for the catch.

test/api/route_test.js
jonasfj: first test returns the querystring... but it's not very interactive... second test returns empty as no querystring is provided...

jonasfj: I should probably use a slugid to ensure that it's unique...

petemoore:
Thanks Jonas. Yeah I tend to think it is overkill too. However, I think the description idea still might have legs ([here](https://github.com/taskcluster/taskcluster-lib-api/pull/1#issuecomment-160938279))... What do you think?

petemoore:
@jonasfj it looks like http://schemas.taskcluster.net/base/v1/api-reference.json has not picked this change up yet

src/schemas/api-reference.json
```diff
@@ -7,7 +7,7 @@
   "properties": {
     "version": {
       "description":          "API reference version",
-      "enum":                 [0, "0.2.0"]
+      "enum":                 [0]
```
petemoore: I'm not sure what the version is for, if we always keep it at 0, also when it is in production. :/

jonasfj: Having the option doesn't cost anything... Using will cost us...

jonasfj:
I'll publish when I publish to npm... Don't worry it'll get there... hs client doesn't have support yet...

PR #2, opened 2015-12-01
Bug 1131057

jonasfj:
This is on top of the previous one I did today.. have a look at the last commit only...

https://bugzilla.mozilla.org/show_bug.cgi?id=1131057

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -138,6 +138,54 @@ var schema = function(validator, options) {
   };
 };
 
+
+/**
+ * Validate query-string against query.
+ *
+ * options:
+ * {
+ *   param1: /.../,               // Reg-exp pattern
+ *   param2(val) { return "..." } // Function, returns message if invalid
+ * }
+ *
+ * Query-string options not specified in options will not be allowed. But it's
+ * optional if a request carries any query-string parameters at all.
+ */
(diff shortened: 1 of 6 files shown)


src/api.js
```diff
@@ -102,6 +102,12 @@ var schema = function(validator, options) {
   return function(req, res, next) {
     // If input schema is defined we need to validate the input
     if (options.input !== undefined && !options.skipInputValidation) {
+      if (req.headers['content-type'] !== 'application/json') {
+        return res.status(400).json({
+          message: "Payload must be JSON with content-type: application/json " +
+                   "got content-type: " + (req.headers['content-type'] || null),
+        });
+      }
```
jhford: I like this.  I also like that the error message is not crappy!  Do we need to worry about case of the headers?  For the s3-copy-proxy stuff, I'm using a library called Caseless to help with headers... Maybe we need something like that here?

jhford:
Thanks!  I like this!

src/api.js
jonasfj: node always provides headers in lower case, see: https://nodejs.org/api/http.html#http_message_headers

It's really nice that they go things like that right at platform level :)

PR #3, opened 2015-12-17
Formalizing error handling

jonasfj:
Hi, @jhford I wanted to get your thoughts on this before I take it any further..

I've already over thought this 2-3 times... Initially thinking I wanted a library to wrap raven so we could move away from sentry if we ever wanted to... But I'm increasingly thinking that would be overkill.
(sentry being open source we can just host it on docker if the hosted solution doesn't work out).

Note. I still need tests and I need to use the `res.reportError` method when schema validation fails, etc, like documenting errorCodes in the JSON reference...
I just wanted to get a sanity check of what I have so far... In practice it'll look like:
```js
let api = new base.API({
  ...,
  errorCodes: {
    // List of allowed error codes and their HTTP status code
    MyCustomError: 400
  }
});

api.declare({....}, function(req, res) {
  res.reportError('MyCustomError', "Something went wrong in this list: {{data}}, good luck", {
    data: [1,2,3,4,5,{}]
  });
})
```

Basically the third parameter for `reportError` is a `details` object that can be referenced using `{{key}}` in the second parameter `message`. Only error codes listened in `errors.js` and `new base.API(..` will be allowed, so they'll be documented on a per-component level. This is to just keep it simple, we could define errorCodes per api entry, but it's just a lot of work and typing, and will take up a lot of unnecessary space in the documentation.


jhford:
Since we're declaring the errors that *could* happen early enough, why not save a little typing and make the report methods be like:
```javascript
res.reportMyCustomError('{{data}}', {data: [1,2,3})
```
We should have an assertion that the error names follow some sort of pattern like `[A-Z][a-zA-Z0-9]*` anyway, and if we made it a valid JS identifier, it'd be nicer to use, imo.

src/errors.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * The `method` is the name of the API method, `errorCodes` is a mapping from
+ * allowed error codes to HTTP status codes, and `raven` is an instance of
+ * `raven.Client` from the `raven` npm module.
+ */
+let BuildReportErrorMethod = (method, errorCodes, ravent = null) => {
+  return (req, res, next) => {
+    res.reportError = (code, message, details = {}) => {
+      let status = errorCodes[code];
+      if (status === undefined || typeof(message) !== 'string') {
```
jhford: message not being a string doesn't totally mean that it's missing... could we try JSON.stringify'ing it and stick that in?  Ugly, but I'd prefer an ugly message to a generic one :/

src/api.js
```diff
@@ -1446,6 +1453,7 @@ API.prototype.declare = function(options, handler) {
  *     accessToken:       '...'   // Access token for clientId
  *     // Client must have the 'auth:credentials' scope.
  *   },
+ *   raven:               null,   // raven.Client instance for error reporting
```
jhford: maybe phrase "// optional raven.Client instance for error reporting"

jonasfj:
> We should have an assertion that the error names follow some sort of pattern like `[A-Z][a-zA-Z0-9]*` anyway, and if we made it a valid JS identifier, it'd be nicer to use, imo.

Yeah, we should definitely have a regexp for allowed error names, that's a great idea.

>```javascript
res.reportMyCustomError('{{data}}', {data: [1,2,3})
```
I briefly considered this, and it certainly looks prettier. The downside is that a spelling error becomes a javascript error `undefined is not a function` which isn't very useful. If we had static type-safety doing this would be cool, but we don't... So I'm tempted to say that `res.reportError(code, message, details)` is better, because an invalid code can still trigger a 500 error message that contains some useful information and it's a lot easier to track down the error.

(that said, I agree using a typed name would look better in the code).

File: src/api.js (hunk 2 above)
Comment (jonasfj) [inline src/api.js]: but that would break my line width :)

Okay, probably we should make that clear...

File: src/errors.js (hunk 1 above)
Comment (jonasfj) [inline src/errors.js]: If message is missing it will be included in the 500 error:
```js
 message = 'Internal error, unknown error code: ' + code + '\n' +
              (message || 'Missing message!');
```

jonasfj:
I had to retrigger the test, I suspect it was an NPM error... from looking at the log, it seemed like something it was building failed.

It didn't get to the stage of running the tests or even compiling them, so I think the intermittence is on npm/travis...

jhford:
This looks good to me!  I'm looking forward to using it.

PR #4, opened 2016-01-08
req.scopes()

jonasfj:
@gregarndt needs this method to forward some authority in mozilla-taskcluster when doing retriggers on behalf of users in a background job...

Generally, I'm not a huge fan of this, because I would like a future where we sent the `authorization` header and required scopes to `auth.taskcluster.net` for validation... Rather than getting the list of authorized scopes in return from `auth.taskcluster.net`...  This is why I made the function async, so that we can still support it in a future where `req.satisfies()` doesn't already have the set of scopes, but instead sends the scope-set to `auth.taskcluster.net` for validation...


jonasfj:
@jhford, this is based on https://github.com/taskcluster/taskcluster-lib-api/pull/3, so you might want to review and merge that first... Overall this is a pretty small change.

PR #6, opened 2016-01-25
Fix bug 1234929

jonasfj:
I think this will allow proper expansion of scopes in auth...


djmitche:
:+1:

jonasfj:
@djmitche, new version is out... Enjoy `0.9.3`.

PR #7, opened 2016-01-26
Removed property 'query' from requiredProperties

petemoore:
For API references, `query` is an optional property, only required when
an API endpoint accepts querystring parameters. Most of our API
endpoints do not accept querystring parameters, and most endpoints
currently do not include it in their definition.  However, currently it
is listed as a required property, which means our references are now not
conformant with the schema. This commit removes `query` from the
requiredProperties list so that it can be omitted in the majority of
cases where it is not needed, and fixes the references to be compliant
with the schema again, as currently most are invalid.

### DIFF
--- a/src/schemas/api-reference.json
+++ b/src/schemas/api-reference.json
@@ -118,7 +118,7 @@
         },
         "additionalProperties":   false,
         "required": [
-          "type", "method", "route", "args", "query",
+          "type", "method", "route", "args",
           "name", "stability", "title", "description"
         ]
       }


petemoore:
@jonasfj FYI

jonasfj:
@petemoore, I think this is wrong..

You shouldn't do validation when reading the data... so having query missing some places is not a problem... just implement your tool to support that.

I suspect this is because you generate the object used to read this from the reference schema... Which I think is an anti-pattern, at least for references... As we do that schema for docs mostly...Anyways, we can unmerge later...

petemoore:
@jonasfj We discussed this on IRC shortly after your comment, and the following I provide for the purposes of openness. We remain in disagreement.

The references define a contract between provider and consumer, and we even [point consumers to the references on our website](http://docs.taskcluster.net/tools/references/). Not only this, we advise that these references are useful for the "automatic generation of client libraries". By making `query` a `requiredProperty`, most of our reference schemas are now invalid against the schema they themselves define as being the schema they conform to. We should not have reference schemas declaring conformance to a json schema that they do not satisfy.

Note, this would have also been solved if we versioned our schemas, but we decided not to version our schemas, and to accept the cost of keeping the schemas backwardly compatible until referencing schemas are updated.

jonasfj:
Okay, you win :)

PR #8, opened 2016-01-29
[WIP] support temp creds with clientId != ext.certificate.issuer

djmitche:


djmitche:
I'm going to do some refactoring of tc-auth, tc-lib-api, and tc-lib-testing first..

PR #9, opened 2016-02-03
move local signature validator to auth

djmitche:
This also removed a bunch of dead code for the old (`auth:credentials`) way of authenticating.

See https://github.com/taskcluster/taskcluster-auth/pull/39


djmitche:
I think this should ship with a major version bump for the library, so that existing services can keep using the old library.

jonasfj:
So you'll move `test/auth_test.js` into the `tc-auth`, right?

djmitche:
Hm, `test/auth_test.js` is actually testing the mock auth server, which I discovered when introducing errors into the real signature validator did not cause any test failures.  But, yes, (a) that will be a good starting point for auth tests and (b) I should remove it from tc-lib-api and start using fakeauth there.

jonasfj:
`r+`, but just so you know...

Porting all tests to use fakeAuth is a non-trivial effort. There is a lot of components to move.

djmitche:
Yeah, we can take our time porting to fakeauth.  Per component, it's not that much work.

djmitche:
OK, I stripped down the tests a bit (so they focus on stuff the API is doing, around scope validation) and refactored to use fakeauth.

jonasfj:
Looks okay.. `r+` stands.

PR #11, opened 2016-02-09
Expose calling clientId with req.clientId

djmitche:
This will require https://github.com/taskcluster/taskcluster-lib-testing/pull/6 along with a version bump to 0.10.1.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -298,10 +298,15 @@ var createRemoteSignatureValidator = function(options) {
  * `'superuser'`. If the client has pattern "service:*" this will match any
  * scope that starts with "service:" as is the case in the example above.
  *
- * This also adds a method `req.satisfies(scopes, noReply)` to the `request`
- * object. Calling this method with a set of scopes-sets return `true` if the
- * client satisfies one of the scopes-sets. If the client does not satisfy one
- * of the scopes-sets it returns `false` and sends an error message unless
+ * The request grows the following properties:
+ *
+ *  * `req.satisfies(scopesets, noReply)`
+ *  * `req.scopes()`
+ *  * `req.clientId`
+ *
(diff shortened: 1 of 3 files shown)


src/api.js
```diff
@@ -373,16 +382,42 @@ var remoteAuthentication = function(options, entry) {
...
         }
-        return Promise.resolve([]);
+        return Promise.resolve(result.scopes || []);
       };
 
+      if (result.status === 'auth-success') {
+        req.clientId = result.clientId || '(unknown)';
+      } else {
+        req.clientId = '(' + result.status + ')';
+      }
```
jonasfj: Can we make this a function that returns a promise...
I did this with `scopes()` specifically, so that we could, in some future, make `authenticate` something we do when it's requested... Rather than doing it up-front for all API calls.

jonasfj: just `req.clientId = () => Promise.resolve(result.clientId);` 

Hmm... maybe `req.clientId = () => Promise.reject(new Error(result.message));` is a lot safer than returning "unknown".

djmitche: I can make this an async function like scopes, but it seems superfluous.

I want to always return a clientId, since this is used for auditing and logging -- throwing an error would probably mask another, real problem in this case.

djmitche:
OK, updated to make it an async function

jonasfj:
`r+`

At some point I want to also make `req.satisfies` async, but that involves porting a lot of code...
When we do, we might want to refactor it such that there is a declarative way to go from parameters in scopes to scopes without parameters... so we don't have to do the `if (!req.satisfies` dance... This is just a good preparation for that day.

djmitche:
Thanks -- I bumped tc-lib-api to 0.10.2 and published.

PR #12, opened 2016-02-11
Use hasOwnProperty, and JSON.stringify non-strings

jonasfj:
@imbstack, let me know if you think this is sane...

I suspect it was a bug in the original code.

Oh, if you like let's bump version and bump it in tc-base too.. So we can get this everywhere...


imbstack:
r+

PR #13, opened 2016-02-11
Authentication failed

jonasfj:
@djmitche and @imbstack, do you agree...

I've already made some unit tests that checks `err.code === 'AuthenticationFailed'`.
But I don't think we have too many... I suspect we should do this now, before it's too late.

I think authorization failure will always use the: `InsufficientScopes` error code.
And if that's not applicable, we should introduce: `AuthorizationFailed: 403`

Agree? (I always get this wrong so I better ask two people)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "taskcluster-lib-api",
-  "version": "0.10.3",
+  "version": "0.10.5",
   "author": "Jonas Finnemann Jensen <jopsen@gmail.com>",
   "description": "taskcluster-lib-api",
   "license": "MPL-2.0",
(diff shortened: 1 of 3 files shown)


djmitche:
https://httpstatuses.com/401

If you send a 401, you'll also need to send a `WWW-Authenticate: Hawk` header.

jonasfj:
> If you send a 401, you'll also need to send a WWW-Authenticate: Hawk header.

Oh, no someone read the docs... I had hoped no-one would ever have noticed my crime here :)
I guess we have to serious consider sending that header though... it should really cost us anything...

djmitche:
# 06
The effectiveness of the Internet as a public resource depends upon interoperability (protocols, data formats, content), innovation and decentralized participation worldwide.
---

I'd say that standards count as a top-level priority for us :)

djmitche:
+1

PR #14, opened 2016-03-15
Bug 1255821: Generate valid clientIds for exceptional cases

djmitche:
The invalid clientIds are causing validation errors in testAuthenticate responses.


jonasfj:
Nice... I can totally see corner cases where we use this blindly...

PR #15, opened 2016-03-16
Use tc-lib-validate in place of schema-validator-publisher

imbstack:
This is an api breaking change, but introduces some nice errors in place of the opaque ones that the old library had.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "taskcluster-client": "0.23.4",
     "taskcluster-lib-scopes": "^0.8.8",
     "taskcluster-lib-stats": "^0.8.7",
-    "taskcluster-lib-validate": "^0.4.2",
     "type-is": "^1.6.10",
     "uuid": "^2.0.1"
   },
@@ -54,6 +53,7 @@
     "taskcluster-lib-config": "^0.8.8",
     "taskcluster-lib-testing": "^0.10.1",
     "taskcluster-lib-app": "^0.8.9",
+    "taskcluster-lib-validate": "^0.4.2",
     "tc-rules": "^2.0.0"
   },
   "main": "./lib/api.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "taskcluster-lib-api",
-  "version": "0.10.5",
+  "version": "0.11.0",
   "author": "Jonas Finnemann Jensen <jopsen@gmail.com>",
   "description": "taskcluster-lib-api",
   "license": "MPL-2.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "taskcluster-client": "0.23.4",
     "taskcluster-lib-scopes": "^0.8.8",
     "taskcluster-lib-stats": "^0.8.7",
-    "taskcluster-lib-testing": "^0.10.1",
+    "taskcluster-lib-validate": "^0.4.2",
```
jonasfj: isn't this only a devDepenednecy?

src/api.js
```diff
@@ -111,15 +112,15 @@ var schema = function(validator, options) {
...
+      var error = validate(req.body, options.input);
+      if (error) {
         debug("Request payload for %s didn't follow schema %s",
               req.url, options.input);
         return res.reportError(
           'InputValidationError',
           "Request payload must follow the schema: {{schema}}\n" +
-          "Errors: {{errors}}", {
-          errors,
+          "Error: {{error}}", {
```
jonasfj: I suspect we don't need the `Error: ` part... is the `error` variable a full error message with everything needed?

So maybe `res.reportError('InputValidationError', error, {schema: options.input, payload: req.body});` is fine...

src/api.js
```diff
@@ -131,14 +132,14 @@ var schema = function(validator, options) {
...
       // defined, then we have to validate against it...
       if(options.output !== undefined && !options.skipOutputValidation) {
-        var errors = validator.check(json, options.output);
-        if (errors) {
+        var error = validate(json, options.output);
+        if (error) {
           debug("Reply for %s didn't match schema: %s got errors: %j from output: %j",
-                req.url, options.output, errors, json);
+                req.url, options.output, error, json);
           let err = new Error('Output schema validation of ' + options.output);
```
jonasfj: I propose we do: ` new Error('Output schema validation error: ' + error)` as the message is pretty featureful right?

package.json
imbstack: I'm a bit confused. I didn't think this was devDependency only. schema-validator-publisher wasn't. Was that a mistake?

imbstack: It's also imported in `src/api.js` so that seems like a real dependency to me.

jonasfj: nope...
But we get validate passed in now... So we don't need it as a dependency, right?

jonasfj: We also had it passed in before... the only place we used package was in reference() where you use ajv now...

jonasfj:
`r+`, but please make sure the error messages are as good as they can get see my comments..

package.json
imbstack: Ah, yes. Duh.

jonasfj: but not used in src/api.js anymore..

PR #16, opened 2016-03-17
Add InputError

jonasfj:
If we do some manual validation... ie. validation that isn't done by schema validation, regex etc...


imbstack:
r+

PR #17, opened 2016-04-26
Add tc-lib-monitor for stats and error reporting

imbstack:
This adds tc-lib-monitor for monitoring and error reporting while still allowing tc-lib-stats to be used for backwards-compatibility and enable an easy upgrade path. Eventually we can remove tc-lib-stats.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,7 @@
     "superagent-hawk": "0.0.4",
     "superagent-promise": "0.1.2",
     "taskcluster-client": "0.23.4",
-    "taskcluster-lib-monitor": "^0.3.0",
     "taskcluster-lib-scopes": "^0.8.8",
-    "taskcluster-lib-stats": "^0.8.7",
     "type-is": "^1.6.10",
     "uuid": "^2.0.1"
   },
@@ -54,6 +52,7 @@
     "taskcluster-lib-config": "^0.8.8",
     "taskcluster-lib-testing": "^0.10.1",
     "taskcluster-lib-app": "^0.8.9",
+    "taskcluster-lib-monitor": "^0.3.0",
     "taskcluster-lib-validate": "^0.4.2",
     "tc-rules": "^2.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "superagent-hawk": "0.0.4",
     "superagent-promise": "0.1.2",
     "taskcluster-client": "0.23.4",
+    "taskcluster-lib-monitor": "^0.3.0",
     "taskcluster-lib-scopes": "^0.8.8",
     "taskcluster-lib-stats": "^0.8.7",
     "type-is": "^1.6.10",
--- a/src/api.js
+++ b/src/api.js
@@ -13,15 +13,13 @@ var path          = require('path');
 var fs            = require('fs');
 require('superagent-hawk')(require('superagent'));
 var request       = require('superagent-promise');
-var stats         = require('taskcluster-lib-stats');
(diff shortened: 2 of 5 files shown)


src/api.js
```diff
@@ -13,9 +13,9 @@ var path          = require('path');
 var fs            = require('fs');
 require('superagent-hawk')(require('superagent'));
 var request       = require('superagent-promise');
-// Someone should rename utils to scopes... 
-var utils         = require('taskcluster-lib-scopes');
 var stats         = require('taskcluster-lib-stats');
+var scopes        = require('taskcluster-lib-scopes');
+var monitoring    = require('taskcluster-lib-monitor');
```
jonasfj: Why?

src/api.js
```diff
@@ -708,12 +709,17 @@ API.prototype.router = function(options) {
...
+  // Create statistics reporters
   var reporter = null;
   if (options.drain) {
     assert(options.component, "The component must be named in statistics!");
     reporter = series.ResponseTimes.reporter(options.drain);
   }
+  var monitor = null;
+  if (options.monitor) {
+    assert(options.component, "The component must be named for monitoring!");
+    monitor = options.monitor.prefix(options.component + '.api');
```
jonasfj: or, how about we do this inside the loader filer for each component... that way it's more transparent?

I guess we would be repeating it, but then we don't need to pass `options.component` anymore do we?

src/api.js
```diff
@@ -756,11 +762,39 @@ API.prototype.router = function(options) {
         component:  options.component
       }));
     }
+    if (monitor) {
+      middleware.push( (req, res, next) => {
```
jonasfj: Make this a function that returns a  function like all the other middle ware things.... why not?

src/api.js
```diff
@@ -756,11 +762,39 @@ API.prototype.router = function(options) {
...
+          try {
+            // Avoid sending twice
+            if (sent) {
+              return;
+            }
+            sent = true;
+
+            let d = process.hrtime(start);
+
+          for (let statusCode of [res.statusCode, 'all']) {
```
jonasfj: indentation issue...

src/api.js
```diff
@@ -756,11 +762,39 @@ API.prototype.router = function(options) {
...
+            let k = entry.name + '.' + statusCode;
```
jonasfj: I suggest we don't count for each status code... do we really want that...

Maybe we want to count number of successful, failed and 5xx requests, but do we need more?

package.json
```diff
@@ -32,6 +32,7 @@
     "superagent-hawk": "0.0.4",
     "superagent-promise": "0.1.2",
     "taskcluster-client": "0.23.4",
+    "taskcluster-lib-monitor": "^0.3.0",
```
jonasfj: should only be a devdep

package.json
```diff
@@ -32,6 +32,7 @@
...
     "taskcluster-lib-scopes": "^0.8.8",
     "taskcluster-lib-stats": "^0.8.7",
```
jonasfj: should go away right?

imbstack: I kinda want to keep it for backwards-compatibility reasons. It will make upgrading opt-in and easy.

jonasfj: Well just ignore the option if it's there

src/api.js
imbstack: Why scopes or why monitoring? scopes I just changed because I saw a comment above this that said `// Someone should rename utils to scopes... `.

jonasfj: Monitoring

jonasfj: It's not used

PR #18, opened 2016-05-03
Use https:// to refer to schema, references

djmitche:
I *think* this should be harmless since we are now serving the same content on http://schemas.taskcluster.net and https://schemas.taskcluster.net.

But the Sentry errors have me nervous..

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -84,8 +84,8 @@ var parameterValidator = function(options) {
  *
  * options:
  * {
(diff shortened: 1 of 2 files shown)


imbstack:
I _think_ this should work as well. Not sure why it didn't work before though, so I'm just as suspicious of this as you are.

djmitche:
I'm not going to attempt this right now..

PR #32, opened 2016-07-21
Add name tag to internal errors

jonasfj:


imbstack:
This looks good. r+

Note to future self: Jonas and I checked into how raven handles this sort of data and it appears to be how we want it to: https://github.com/getsentry/raven-node/blob/master/lib/client.js#L119-L127

imbstack:
## r+

PR #33, opened 2016-07-28
Fix bug 1184209

jonasfj:


jonasfj:
@petemoore, you might actually want to read this 11 line patch....  have a look...

jonasfj:
We'll have to update clients to support downloading files I think... Or downloading a stream... we haven't exactly figured out yet what they API will look like... it probably depends on the language too...

petemoore:
We could keep it simple, and just return one of:

* an http request
* a (signed if required) url
* (if we want to do the retries in the client) an array of bytes

Another alternative is that we require a filename, and we do the retries, and write to that. Not pretty.

I tend to think we should let retries be the consumer's problem - we might have different strategies to what they want.

PR #35, opened 2016-07-29
Use `req.params` by default when deferAuth is false

djmitche:
[Bug 1290549](https://bugzilla.mozilla.org/show_bug.cgi?id=1290549)
identifies at least one place where this is assumed to be the case, and
there are probably more.


jonasfj:
# `r+`

imbstack:
## r+

Jonas has now explained to me how this works, so r+.

djmitche:
Oh, I'll push a semver bugfix bump too

PR #39, opened 2016-09-01
stop using aws-sdk-promise

jhford:
I don't know for sure that this works because the unit tests no longer run for me.  We should consider fixing these unit tests so they're less messy


jhford:
@jonasfj i did check that the potential `.data` property or not is not read, so we shouldn't notice any behaviour changes.

jhford:
@jonasfj Note that this should be a major version release.  Reasoning, this is likely the last library which requires aws-sdk-promise, and because aws-sdk-promise is overwriting the .promise() from the aws-sdk lirbary, if we remove the final aws-sdk-promise thing in here, we're going to change behaviour signficantly.  Even though this patch won't change the behaviour, it *will* change behaviour of downstream users indirectly.

jhford:
the failure for `AssertionError: bad expires` looks like it's completely unrelated to this failure, and rather is a change somewhere in the S3 stack.

jonasfj:
# `r+`

jonasfj:
I think this has landed, closing...

PR #40, opened 2016-09-06
Remove mockauthserver import from lib-testing

imbstack:
This makes the weird import go away. It didn't end up being as hard as I expected, which worries me that tests are just passing for no reason now. If you know of any areas where I missed something, please let me know.

I removed the test "reference from mockAuthServer" entirely because it seemed to just be re-testing something that is testing a bit more simply down below it in the same file. If the extra complexity is worth it, I can go and re-create a mockAuthServer type thing to test publishing on, but I don't believe it is worth it.

The diff here looks a log bigger than it actually is. Most of the "changed" lines are just removing whitespace at the beginning as they became one level less deep inside promises.

bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1298398

### DIFF
--- a/test/publish_test.js
+++ b/test/publish_test.js
@@ -4,58 +4,6 @@ suite("api/publish", function() {
   var aws             = require('aws-sdk');
   var assert          = require('assert');
   var Promise         = require('promise');
-  var mockAuthServer  = require('taskcluster-lib-testing/.test/mockauthserver');
-
-  // Sanity checks for the reference
-  test("reference from mockAuthServer", function() {
-    var reference = mockAuthServer.api.reference({
-      baseUrl:          'http://localhost:23243/v1'
-    });
-    assert(reference.entries, "Missing entries");
-    assert(reference.entries.length > 0, "Has no entries");
-    assert(reference.title, "Missing title");
-  });
-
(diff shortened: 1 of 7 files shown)


test/publish_test.js
```diff
@@ -4,58 +4,6 @@ suite("api/publish", function() {
...
-        Key:        'base/test/api.json'
-      }).promise();
-    }).then(function(res) {
-      var reference = JSON.parse(res.Body);
-      assert(reference.entries, "Missing entries");
-      assert(reference.entries.length > 0, "Has no entries");
-      assert(reference.title, "Missing title");
-    });
-  });
-
```
djmitche: Agreed - 'publish minimal reference', below, does the same thing without co-opting the mockAuthServer's API definition.

djmitche:
Yes, this looks good.  I took a different course of trying to copy the existing MockAuthServer into tc-lib-api and got stuck in some quicksand.  This fix looks good.  Basically you've removed the test of publishing the MockAuthServer (which was done, I assume, "because it's there", and adds no value as you noted), and for the remainder ported things from using MockAuthServer to FakeAuth.

PR #41, opened 2016-10-03
Bug 1302753: log all successful authorizations

djmitche:
Result:
```
Authorized rockstar for GET access to /test-static-scope
```


jonasfj:
Ie the api signature name... Like createTask

djmitche:
I don't think that value is easily accessible here.  This code is in the
remoteAuthentication implementation.  I think this is good enough?

jonasfj:
Okay, then send it off..

Den 4. okt. 2016 10.34 AM skrev "Dustin J. Mitchell" <
notifications@github.com>:

PR #43, opened 2016-10-14
One ping

owlishDeveloper:


src/api.js
```diff
@@ -670,7 +687,7 @@ API.prototype.declare = function(options, handler) {
   if (options.output && options.output !== 'blob') {
     options.output = this._options.schemaPrefix + options.output;
   }
-  this._entries.push(options);
+  this._entries.unshift(options);
```
djmitche: This will end up reversing the whole list of endpoints, right?  I'd rather see a way of including ping at the end while maintaining the order of the others.

src/api.js
```diff
@@ -555,7 +555,24 @@ var API = function(options) {
...
+          'Respond without doing anything.',
+          'This endpoint is used to check that the service is up.'
+        ].join('\n'),
+      handler: function(req, res) {
+          res.status(200).json({
+            alive:    true,
+            uptime:   process.uptime()
+            });
+        }
+     }];
```
djmitche: Will this cause an error, or list the ping endpoint twice, for services that already define such and endpoint?

owlishDeveloper: I was planning to eliminate the other ping declarations. I'm not sure though. What do you think?

src/api.js
owlishDeveloper: Yeah, it sure will. Ok, I'll think about it!

owlishDeveloper: Made some changes, as far as my understanding of Jonas's suggestion ("probably append it to _entries whenever we read the property") goes. What do you think?

src/api.js
```diff
@@ -833,6 +853,7 @@ API.prototype.router = function(options) {
 API.prototype.reference = function(options) {
   assert(options,         "Options is required");
   assert(options.baseUrl, "A 'baseUrl' must be provided");
+  this._entries.push(ping);
```
djmitche: It looks like this is getting pushed twice now.  I don't think `this._entries` should be modified by functions that don't say they do so (for example, it's reasonable to think I could call `api.reference()` twice and get the same answer both times).

In both of these cases, could you do something like `_.concat(this._entries, [ping]).forEach`.  That joins the two arrays but doesn't modify `this._entries` in the process.

owlishDeveloper: Of course! Thank you so much, Dustin!

imbstack:
It appears the tests are failing. https://travis-ci.org/taskcluster/taskcluster-lib-api/jobs/168687831#L1791

Does that seem related to the changes here?

owlishDeveloper:
Yeah, it's failing because of my changes. That test declares an API and adds it to _entries. After that, it checks how many entries are there in the _entries, and if the number is not 1, it fails. It's not 1 now because I'm having a default entry here. I can try to change the test so it would assert 2 entries instead of 1 to see what happens.

owlishDeveloper:
See, it's not failing now.

djmitche:
Regarding using lodash -- you know better than I actually!  Either one is fine, but I didn't realize that `Array.prototype.concat` was available here (since it's ES5).  I don't really understand the different language versions very well.  At any rate, this is fine as-is.

djmitche:
Brian, should we make a new release?  Probably with a minor version bump?

owlishDeveloper:
Er, just to check: is this ping already in use? Because I tried deleting declaration of ping from taskcluster.auth API, and the tests are failing where they seemingly shouldn't, so I decided to check :)

djmitche:
We do use the ping endpoint for monitoring in production, but I don't know
if we have tests for it.

Dustin

PR #44, opened 2016-10-31
Only log authorizations in production mode

djmitche:
This prevents authorizations from polluting the console output of test
runs of services using this library.

As suggested in https://github.com/taskcluster/taskcluster-pulse/pull/35


jonasfj:
Oh tests fail because it's on your repo...

PR #48, opened 2017-01-09
optionally write an api reference file for local debugging

jhford:


jhford:
I cannot run the tests locally, but baring simple problems, this is the approach i think we should take.

jonasfj:
This is why I wanted to do it tc-lib-docs so we don't have this code both places... In fact we also need it into pulse-publisher.
Let's discuss it in: https://github.com/taskcluster/taskcluster-lib-validate/pull/31

I assume none of this urgent.

jhford:
I really don't understand why we need a second library to essentially do `fs.writeFileSync('api-reference.json', JSON.stringify(X))`.  This is a simple debugging tool, nothing to do with docs.  Even if lib-docs could do this, I really don't think that this simple behaviour belongs there.

src/api.js
```diff
@@ -950,6 +950,28 @@ API.prototype.publish = function(options) {
...
+
+  let content = JSON.stringify(this.reference(options), undefined, 2);
+  
+  return new Promise((resolve, reject) => {
+    fs.writeFile('api-reference.json', content, err => {
+      if (err) {
+        reject(err);
+      }
+      resolve();
+    })
```
jonasfj: Semi colon missing :)

djmitche:
The new `write` support in taskcluster-lib-docs supports this same functionality.  Services now have a `writeDocs` component that will write out docs, refs, and schemas to a directory of your choosing.

PR #50, opened 2017-01-20
Vastly improve documentation in README.md

djmitche:


djmitche:
This is part of a concerted effort to put full library docs in the README.  The README appears on github, on npmjs.org, and on docs.taskcluster.net, so it makes sense to put the canonical docs there.

src/api.js
```diff
@@ -573,94 +554,21 @@ var API = function(options) {
...
- * }
- *
- * The handler parameter is a normal connect/express request handler, it should
- * return JSON replies with `request.reply(json)` and errors with
- * `request.json(code, json)`, as `request.reply` may be validated against the
- * declared output schema.
- *
- * **Note** the handler may return a promise, if this promise fails we will
- * log the error and return an error message. If the promise is successful,
- * nothing happens.
```
jonasfj: I have mixed feelings about removing these docs from the source... It makes it really hard to write code.
I use this 100% when editing the code... Also If removing it I would like this example and description like format to maintained in the readme.

Or we should make the readme docs a table with: `name`, `type`, `example` and `description`... But I don't see how readme can nicely document things like `query`... This doc format is extremely concise and quick to read, can we please maintain it?

djmitche: I can leave the changes in place, if you'd like.  I kinda don't like the format, but we can both win :)

PR #51, opened 2017-01-25
Bug 1330978: reduce redundancy in API error messages

djmitche:
The tests have some nice summaries of the error payloads at each stage of this.

My thinking is to get this in place, then do some work on the tools site to just show message by default, with an 'advanced' button to show the request info.  It occurs to me, maybe that request info should include the service name somehow, to help the user understand which call failed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,8 +273,8 @@ into `details`, with the result being JSON-encoded.  For example:
 ```js
 res.reportError(
   'TooManyFoos',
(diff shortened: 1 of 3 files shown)


djmitche:
Hmm, and query should probably be in there with params.  Let me know what you think and I'll update.

jonasfj:
Feel free to publish a new version

djmitche:
v3.1.0

PR #52, opened 2017-02-17
Don't log authorizations except in production

djmitche:
This is mostly just about my compulsion to get nice, clean `npm test` output :)

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,7 +1,7 @@
 "use strict";
 
 var express       = require('express');
-var debug         = require('debug')('base:api');
+var Debug         = require('debug');
 var Promise       = require('promise');
 var uuid          = require('uuid');
 var hawk          = require('hawk');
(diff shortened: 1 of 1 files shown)


src/api.js
```diff
@@ -42,6 +42,13 @@ var ping = {
   }
 };
 
+var debug = Debug('api');
+
+/* In production, log authorizations so they are included in papertrail regardless of
+ * DEBUG settings; otherwise, log with debug
+ */
+var authLog = process.env.NODE_ENV === 'production' ? console.log : Debug('api.authz');
```
jonasfj: Maybe this is a nit, I know it's not required if I test it in node... but how about:
```js
var authLog = (...args) => console.log(args...);
if process.env.NODE_ENV === 'production') {
  var authLog = Debug('api.authz');
}
```
I'm slightly concern about not binding `console.log` to `console`.
`console.log.bind(console)` also works... Also I like not having it in one line, but that's a nit.

djmitche: Makes sense.

djmitche:
pushed to npm as 3.1.1

PR #53, opened 2017-02-22
Allow an endpoint to be declared as noPublish

jhford:
This means that the API will be ignored for publishing reference files, and thus invisible in documentation and client libraries, but will still be set up in the Express routing framework.  This means that we can support old APIs while not advertising their existance


jhford:
I did #54 in my branch locally but didn't push

PR #56, opened 2017-03-16
Add dry run support

jhford:
I think this is the code we'd need to add for dry run support.  I haven't written unit tests for it yet

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -146,9 +146,20 @@ var schema = function(validate, options) {
         {schema: options.input});
       }
     }
+
+    // We want to ensure that a dry run success returns a non-200 status
+    // code because we want dry runs to be explicitly handled.
+    res.reportDryRunSuccess = (message) => {
+      let dryRunId = uuid.v4();
+      res.status(418).json({
+        code: 418,
+        message: message || 'dry run success',
+      });
+    };
+
     // Add a reply method sending JSON replies, this will always reply with HTTP
     // code 200... errors should be sent with res.json(code, json)
-    res.reply = function(json) {
+    res.reply = function(json, dryRunMessage) {
       // If we're supposed to validate outgoing messages and output schema is
       // defined, then we have to validate against it...
       if(options.output !== undefined && !options.skipOutputValidation &&
@@ -164,7 +175,11 @@ var schema = function(validate, options) {
         }
       }
       // If JSON was valid or validation was skipped then reply with 200 OK
-      res.status(200).json(json);
+      if (req.dryRun) {
+        res.reportDryRunSuccess(dryRunMessage);
+      } else {
+        res.status(200).json(json);
+      }
     };
 
     // Call next piece of middleware, typically the handler...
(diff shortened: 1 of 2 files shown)


src/api.js
```diff
@@ -164,7 +175,11 @@ var schema = function(validate, options) {
         }
       }
       // If JSON was valid or validation was skipped then reply with 200 OK
-      res.status(200).json(json);
+      if (req.dryRun) {
```
jhford: We could also interscept the other response-ending methods ( e.g. `.json` to do this check as well.

jonasfj:
I have mixed feelings about adding dry-run support.

Serverside it's a quite a bit of extra logic and special cases to maintain. Depending on how we want to do the dry-run.
 * Do you want to check that `task.dependencies` exist?
 * Do you want to check that scopes are granted?
 * Do you want to check if `taskGroupId` is already owned by another `schedulerId`?
 * Do you want to check if `taskId` is already in use?
Most of our input checks are just the JSON schema, additional input tests which are written manually would be hard to maintain. Furthermore, those tests may sometimes be undesirable, for example checking if `task.dependencies` exist makes dry-run unusable for in-tree stuff.

I see two things we could do:
1) Add a generic dry-run option in tc-lib-api, it would only validate declared scopes and declared schema, nothing more. So no custom logic in all the applications.
2) Add a dry-run option in client libraries by validating against declared input schemas.

Requiring every single API end-point to handle cases where `req.dryrun == true` adds a lot of complexity to our services. Increases maintainability and even then we still don't really know which properties we want to validate during a dryrun validation. Hence, I think it's very tempting to say that dryruns should be JSON schema validation only, well, maybe declared scopes too if possible.

Right now we use `deferAuth` a lot! because we inject patterns into our scopes. But @imbstack, @djmitche and I have been talking about changing `auth.authenticateHawk` in such a way that the request we send to `auth` contains the list of scopes we want and returns true/false, rather than just returning the list of scopes the client has. This will likely require some refactoring of how patterns are injected into scope patterns. Hence, it's not unthinkable that a future refactoring of tc-lib-api will have:
```js
api.declare({
   ....
   scopes: ['service:get:<key>'...],
   scopeParams: (req) => return {key: req.body.pathToFile},
})
```
So that instead of doing an imperative `req.satisfies` call we declare a function `scopeParams` which figures out what scopes a request needs, without side-effects. Then we would be in a good position to have generic dryrun support that also checks scopes containing patterns.

So maybe we do a generic dryrun that for now just validates JSON schemas. If you are sure we want to validate more for certain methods... Maybe `api.declare({dryHandler: ...` is an options. So for particularly important API methods we can add checks beyond JSON schema validation.

petemoore:
@imbstack convinced me now (it took a while 😜 ) that dry runs could dirty the waters, so I think I'm going to step back from this now. But I'm extremely grateful you stepped in so quickly to help out, thanks @jhford.

jhford:
@jonasfj then that means we need to validate JSON schema in the client library, which is something you've been opposed to in the past.

I disagree that dry run support is a sign that our system is broken.  I actually see being able to implement a generic dry run ability as a sign that we've designed a sufficiently declarative API that we can do it.  We could implement it with checking just the declarative properties by default and having api handlers opt-in to receive the ability to do an application-logic dry-run.

PR #57, opened 2017-03-22
Stop having superagent as a direct dependency since we do not use it

jhford:
This causes an old version of Superagent to get installed in all depending-on-this projects.  Since we're not actually making requests as part of this library, we shouldn't need to isntall it in this library


djmitche:
@jhford do you want to rebase (or re-construct) this?  I'm not sure why it never landed..

jhford:
Looks like this has landed through other commits and is no longer needed

PR #58, opened 2017-03-31
Move authlog to use monitor audit log

imbstack:
As we discussed on irc. This moves the writing to audit logs to lib-api.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -44,14 +44,6 @@ var ping = {
 
 var debug = Debug('api');
 
-/* In production, log authorizations so they are included in papertrail regardless of
- * DEBUG settings; otherwise, log with debug
- */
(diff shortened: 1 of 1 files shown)


src/api.js
```diff
@@ -361,7 +353,21 @@ var createRemoteSignatureValidator = function(options) {
...
+   */
+  let authLog;
+  if (options.monitor) {
+    authLog = options.monitor.log;
+  } else if (process.env.NODE_ENV !== 'production') {
+    let adb = Debug('api.authz');
+    authLog = (obj) => adb(JSON.stringify(obj));
+  } else {
+    authLog = (obj) => console.log(JSON.stringify(obj));
+  }
```
djmitche: I like this!

PR #59, opened 2017-04-03
Add a noAuth option

imbstack:
Does this generally make sense for endpoints like authenticateHawk?

https://bugzilla.mozilla.org/show_bug.cgi?id=1352581

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ To declare an API method, call `api.declare(options, handler)` with the followin
  * `scopes` - scopes required for this API endpoint, in disjunctive normal form (see below)
  * `deferAuth` - if true, authentication will not be checked automatically before the handler
    is invoked.  In this case, the handler must call `req.satisfies()`; see below.
+ * `noAuth` - if true, this endpoint will not be authenticated and clients should not send
+   headers or bewit
  * `stability` - API stability level, defaulting to experimental (see below)
  * `input` - the schema against which the input payload will be validated
  * `skipInputValidation` - if true, don't do input validation (but include the schema in documentation)
--- a/src/api.js
+++ b/src/api.js
@@ -382,8 +382,8 @@ var remoteAuthentication = function(options, entry) {
     }
 
     // If no authentication is provided, we just return valid with zero scopes
-    if ((!req.query || !req.query.bewit) &&
-        (!req.headers || !req.headers.authorization)) {
+    if (entry.noAuth || ((!req.query || !req.query.bewit) &&
+        (!req.headers || !req.headers.authorization))) {
       return Promise.resolve({
         status: 'no-auth',
         scheme: 'none',
@@ -679,6 +679,10 @@ API.prototype.declare = function(options, handler) {
   if (!options.stability) {
     options.stability = stability.experimental;
   }
+  if (!options.noAuth) {
+    options.noAuth = false;
+  }
+  assert(!options.scopes || !options.noAuth, 'Cannot specify noAuth if the function has scopes');
   assert(STABILITY_LEVELS.indexOf(options.stability) !== -1,
          "options.stability must be a valid stability-level, " +
          "see base.API.stability for valid options");
--- a/test/auth_test.js
+++ b/test/auth_test.js
@@ -92,6 +92,20 @@ suite("api/auth", function() {
     }
   });
 
+  // Declare a method we can test no auth
+  api.declare({
+    method:       'get',
+    route:        '/test-no-auth',
+    name:         'testNoAuth',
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -97,6 +97,8 @@ To declare an API method, call `api.declare(options, handler)` with the followin
  * `scopes` - scopes required for this API endpoint, in disjunctive normal form (see below)
  * `deferAuth` - if true, authentication will not be checked automatically before the handler
    is invoked.  In this case, the handler must call `req.satisfies()`; see below.
+ * `noAuth` - if true, this endpoint will not be authenticated and clients should not send
```
jonasfj: How about `public: true`?

can you explain a bit more why this is necessary?
`deferAuth` should generally be enough... since there is no reason to do auth until `satisfies()` is called..
hmm... I guess the problem is that `satisfies()` is synchronous?

jonasfj:
@imbstack,
Perhaps it would be better to design it as:
```js
api.declare({
  ...,
  scopes: [...], // scope patterns for documentation
  auth: 'optional|required',
}, async function(req, res) {

  // construct scopes and ask for authorization
  if (await req.authorize([['scope-A', 'scope-B'], ['scope-C']])) {
    return;
  }

  // .... whatever
});
```

And then we change behavior such that:
1. If `auth: undefined` then `scopes: undefined` must also be the case, and `await req.authorize(...)` may not be called. If `req.authorize` is called it causes an internal server error.
2. If `auth: 'optional'` then `scopes` must be defined, and `await req.authorize(...)` may be called.
3. If `auth: 'required'` then `scopes` must be defined, and `await req.authorize(...)` **must** be called. If `req.reply(...)` is called because `req.authorize(..)` then it causes an internal server error...  

This way we **always** `deferAuth` if `auth` is specified... (we deprecated `deferAuth` and `req.satisfied`). We also get a security improvement in that if we specify `scopes: [...]` we must specify either `auth: 'optional'` or `auth: 'required'`. And if we specify `auth: 'required'` and we forget to call `await req.authorize(...)` on some code path we get an internal error in sentry tell us that we have security bug... (we've had these before where we forgot to call `req.satisfies` and had `deferAuth: true`).

One downside I see with this design is that we can accidentally do: `if (!req.authorize(...)` instead of `if (!await req.authorize(...)) {`... Ie. forgetting to the `await`...  Maybe we say that `await req.authorize(..)` returns `true` if it fails to authorize (that's a little non-intuitive), but saves us when you forget to `await`...
Or we renamed `if (await req.authFail(...)) { return; }`... Or we make `req.authorize` throw a special error instance that we catch in `tc-lib-api` such that you only have to write `await req.authorize(...)` no need for `if(...) {return;}` and then `tc-lib-api`  just knows that when it gets this special error... it should handle it as a special case.. that might be the elegant solution..

```js
let API = require('taskcluster-lib-api');
let api = new API(...);

api.declare({
  ...,
  scopes: [...], // scope patterns for documentation
  auth: 'optional|required',
}, async function(req, res) {

  // construct scopes and ask for authorization
  // if authorization fails, this will throw a special error object that tc-lib-api handles
  // and turns into a 401 or 403 error... instead of an 500 error as we normally do for
  // unhandled exceptions...
  await req.authorize([['scope-A', 'scope-B'], ['scope-C']]);
  // IMO, getting rid of the: if (...) {return;}  can save us a lot of bugs and pain...

  // In fact we can probably allow things like:
  throw new API.AuthorizationError([scopes....]); // if you want to explicitly cause a 403

  // .... whatever
});
```

-----
My motivation here is to get the design we want long-term, elegant, simple and reasonably free from footguns... Then we can start migrating towards it slowly... That's better than adding another hack to what we currently have... and not much more work.

djmitche:
What if we merge noAuth now, and use it only for authenticateHawk, then move discussion of the new auth approach to an RFC?

authenticateHawk is effectively doubling our Hawk, so the earlier in the request process we can stop calculating Hawk headers for it, the better.

jonasfj:
`authenticateHawk` as called by the API shouldn't make signed request:
https://github.com/taskcluster/taskcluster-lib-api/blob/87b373e7de94e9f55c16602009071483dba61005/src/api.js#L286-L290

If it does it's because it finds default credentials for environment variables:
https://github.com/taskcluster/taskcluster-client/blob/master/lib/client.js#L42-L46

Maybe a trick is specify `credentials: {}` in:
https://github.com/taskcluster/taskcluster-lib-api/blob/87b373e7de94e9f55c16602009071483dba61005/src/api.js#L287

If this is strictly about avoiding `authorization` headers for `authenticateHawk` calls...

PR #62, opened 2017-09-22
Bug 1338047 - Migrate all taskcluster-lib-* libraries to use import/export syntax

ydidwania:
Converted all the files to use import/export syntax , except [test/validate_test.js](https://github.com/taskcluster/taskcluster-lib-api/blob/master/test/validate_test.js).
Following error on migrating the rest of the file.
```
1) api/auth request with static scope:
     TypeError: _superagentPromise2.default.get(...).hawk is not a function
2) api/auth request with static scope - fail no scope:
     TypeError: _superagentPromise2.default.get(...).hawk is not a function
3) api/auth static-scope with authorizedScopes:
     TypeError: _superagentPromise2.default.get(...).hawk is not a function
4) api/auth static-scope with authorizedScopes (star):
     TypeError: _superagentPromise2.default.get(...).hawk is not a function
```
...continues until (8)

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -19,8 +19,8 @@ import cryptiles from 'cryptiles';
 import taskcluster from 'taskcluster-client';
 import Ajv from 'ajv';
 import typeis from 'type-is';
-import 'superagent-hawk';
-import 'superagent';
+
+require('superagent-hawk')(require('superagent'));
   
 
 // Default baseUrl for authentication server
--- a/src/api.js
+++ b/src/api.js
@@ -1,25 +1,27 @@
 "use strict";
 
-var express       = require('express');
-var Debug         = require('debug');
-var Promise       = require('promise');
-var uuid          = require('uuid');
-var hawk          = require('hawk');
-var aws           = require('aws-sdk');
-var assert        = require('assert');
-var _             = require('lodash');
-var bodyParser    = require('body-parser');
-var path          = require('path');
-var fs            = require('fs');
-require('superagent-hawk')(require('superagent'));
-var request       = require('superagent-promise');
-var scopes        = require('taskcluster-lib-scopes');
-var crypto        = require('crypto');
-var cryptiles     = require('cryptiles');
-var taskcluster   = require('taskcluster-client');
-var Ajv           = require('ajv');
-var errors        = require('./errors');
-var typeis        = require('type-is');
+import * as express from 'express';
+import * as _ from 'lodash';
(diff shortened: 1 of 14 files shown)


test/validate_test.js
```diff
@@ -1,11 +1,19 @@
...
-  var subject         = require('../');
-  var helper          = require('./helper');
-  var testing         = require('taskcluster-lib-testing');
+   require('superagent-hawk')(require('superagent'));
+   //var request         = require('superagent-promise');
+   //var assert          = require('assert');
+   //var Promise         = require('promise');
+   //var subject         = require('../');
+   //var helper          = require('./helper');
+   //var testing         = require('taskcluster-lib-testing');
```
djmitche: did these get missed?

jonasfj:
superagent-hawk probably needs some injection... a different approach might be to migrate all usages to `got` which is saner...

src/api.js
```diff
@@ -1,25 +1,27 @@
...
+import fs from 'fs';
+import request from 'superagent-promise';
+import crypto from 'crypto';
+import cryptiles from 'cryptiles';
+import taskcluster from 'taskcluster-client';
+import Ajv from 'ajv';
+import typeis from 'type-is';
+import 'superagent-hawk';
+import 'superagent';
+
```
jonasfj: you are missing `require('superagent-hawk')(require('superagent'));`

jonasfj: you have the imports... but not the expression that does the injection...

test/validate_test.js
ydidwania: I commented them instead of removing just to keep the original imports in sight. 
I will remove the comments once the error is resolved

src/api.js
ydidwania: Yeah. I removed the imports and added `require('superagent-hawk')(require('superagent'));`  in the next commit . Thanks

ydidwania:
In  the commit(221d150) , 
 I added `.test/` to `.gitignore` .
The directory was being created everytime  I ran `npm run test` 
Hope thats fine

ydidwania:
All local tests compete successfully now.

src/api.js
jonasfj: you can still do it with import if you really want... there is two imports and one function invocation in that expression...

IMO; leaving it as require is fine :)

djmitche:
I think we decided not to merge this since we're not using import/export in any of our services, either?  So basically we're (for now) using `require` everywhere..

At any rate, quite stale.

PR #64, opened 2017-11-01
Bug 1408477 - always set cache-control

djmitche:


djmitche:
I'll push this to the upstream repo

PR #65, opened 2017-11-01
Bug 1408477 - always set cache-control

djmitche:


djmitche:
Same as #64, but on the upstream repo so that it can have access to travis secrets.

djmitche:
v3.2.3

PR #66, opened 2017-11-03
Spring Cleaning

imbstack:
Left hawk at <7 since they require a very recent node version to use that as we discussed in irc.


src/api.js
```diff
@@ -417,7 +412,7 @@ var remoteAuthentication = function(options, entry) {
           'sha256',
           req.headers['content-type']
         );
-        if (!cryptiles.fixedTimeComparison(result.hash, hash)) {
+        if (!crypto.timingSafeEqual(result.hash, hash)) {
```
djmitche: nice spot

imbstack: I would love to take credit, but really what happened is that cryptiles removed `fixedTimeComparison` from their most recent version since `timingSafeEqual` was added to the std lib.

PR #68, opened 2018-01-05
Use scope expressions for api authorization

imbstack:
This is for taskcluster/taskcluster-rfcs#23. After this work we should be able to take a few more steps and then get to the world where authorization takes place in taskcluster-auth as well.

This PR presents an implementation of the scope expression templates. I think we're fairly (~78%) confident that this will allow us to cover all of the current use cases of `req.satisfies`. I think my code is a bit messy at the moment, I may take a bit to go back over it and see if I can simplify. Also please feel free to simplify it for me!

In addition if while reviewing you can think of more test cases, please throw them my way and we can add them.

I'm 90% sure you will find huge flaws in my logic because that is the normal case for these reviews, but please be careful and harsh since this is important to get right!

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,23 +137,27 @@ and `workerType`, *or* the caller's scopes satisfy all of
 `queue:define-task:..`, `queue:task-group-id:..`, and `queue:schedule-task:..`.
 
 ```js
-  scopes: {AnyOf: [
-    {AllOf: [
-      'queue:create-task:<provisionerId>/<workerType>',
+  scopes:
+    {AnyOf: [
+      {AllOf: [
+        'queue:create-task:<provisionerId>/<workerType>',
+      ]},
+      {AllOf: [
+        'queue:define-task:<provisionerId>/<workerType>',
+        'queue:task-group-id:<schedulerId>/<taskGroupId>',
+        'queue:schedule-task:<schedulerId>/<taskGroupId>/<taskId>',
+      ]},
     ]},
-    {AllOf: [
-      'queue:define-task:<provisionerId>/<workerType>',
-      'queue:task-group-id:<schedulerId>/<taskGroupId>',
-      'queue:schedule-task:<schedulerId>/<taskGroupId>/<taskId>',
-    ]},
-  ]},
 ```
 
 If scope validation fails, the user is presented with an extensive error
 message indicating the available and required scopes.
 
 Arrays of required scopes can also be templated in the scopes section. When
-this form is used, the parameter must be an array.
+this form is used, the parameter in the `in` section must be an array. The
+value of `each` *must* be a simple template string and cannot be a scope
+expression or another template object. This will not allow recursive
+definitions.
 
 ```js
   scopes: {AllOf: [
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ and `workerType`, *or* the caller's scopes satisfy all of
 If scope validation fails, the user is presented with an extensive error
 message indicating the available and required scopes.
 
-Given an array of parameters, an array of scopes can be templated using `for`. When
+Given an array of parameters, an array of scopes can be templated. When
 this form is used, the parameter in the `in` section must be an array. The
 value of `each` *must* be a simple template string and cannot be a scope
 expression or another template object. This will not allow recursive
@@ -274,9 +274,9 @@ includes some additional security token, its duration should be limited to this
 expiration time to prevent callers from extending their access beyond the
 allowed time.
 
-The `async req.authorize(params, options)` method returns `true` if the client satisfies
-the endpoint's scope expression. If the client does not satisfy the expression, it
-throws an error with the code 'AuthorizationError'. You can catch this if you wish
+The `async req.authorize(params, options)` throws an error with the code
+'AuthorizationError' if the client does not satisfy the scope
+expression in `options.scopes`. You can catch this if you wish
 or let it bubble up and taskcluster-lib-api will return a detailed error message
 to the client.
 
@@ -288,8 +288,7 @@ The AuthorizationError has 3 extra fields to help inspect the results.
 
 The first argument to `req.authorize` must be an object where the keys are parameters
 to the scope expression defined in the method definition. If any parameters are missing
-when you call this, the check will error unless you pass `options.allowLater`. This is
-only for special cases and should be rarely used. The rules for how the scope expression
+when you call this, the check will throw an error. The rules for how the scope expression
 defined in the method is transformed given a set of parameters is described above.
 
 To return a successful result with a JSON body, return `res.reply(result)`.
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ and `workerType`, *or* the caller's scopes satisfy all of
 If scope validation fails, the user is presented with an extensive error
 message indicating the available and required scopes.
 
-Arrays of required scopes can also be templated in the scopes section. When
+Given an array of parameters, an array of scopes can be templated using `for`. When
 this form is used, the parameter in the `in` section must be an array. The
 value of `each` *must* be a simple template string and cannot be a scope
 expression or another template object. This will not allow recursive
@@ -182,11 +182,11 @@ Given the following example:
   scopes: {if: 'private', then: {AllOf: ['foo:bar']}}
 ```
 
-We can call `authorize({private: false})` and the method call will be permitted
+We can call `async req.authorize({private: false})` and the method call will be permitted
 even if the client does not have the scope `foo:bar`.
 
 In addition, parameters can be nested objects and accessed with dotted notation.
-Given parameters `task` that is `{extra: {treeherder: {symbol: 'F'}}}`, we can do
+Given the parameter `task` that is `{extra: {treeherder: {symbol: 'F'}}}`, we can do
 the following:
 
 ```js
@@ -200,8 +200,8 @@ which evaluates to
 ```
 
 When you specify scopes required to access an endpoint, by default all of the params
-specified in the `params` section will be substituted in and the expression satisfaction
-will be checked against the client's scopes. If any of the parameters specified in
+specified in the `params` section of the request will be substituted in and the expression
+satisfaction will be checked against the client's scopes. If any of the parameters specified in
 your scopes are _not_ in the `params`, this satisfaction check will be deferred and the
 endpoint implementation _must_ check for authorization manually as described below. If
 this check does not occur, taskcluster-lib-api will throw an error for the result of the
@@ -274,7 +274,7 @@ includes some additional security token, its duration should be limited to this
 expiration time to prevent callers from extending their access beyond the
 allowed time.
 
-The `req.authorize(params, options)` method returns `true` if the client satisfies
+The `async req.authorize(params, options)` method returns `true` if the client satisfies
 the endpoint's scope expression. If the client does not satisfy the expression, it
 throws an error with the code 'AuthorizationError'. You can catch this if you wish
 or let it bubble up and taskcluster-lib-api will return a detailed error message
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ This will not do any truthiness conversions, you must do that yourself if desire
 Given the following example:
 
 ```js
-  scopes: {AllOf: [{if: 'private', then: 'foo:bar'}]}
+  scopes: {if: 'private', then: {AllOf: ['foo:bar']}}
 ```
 
 We can call `authorize({private: false})` and the method call will be permitted
--- a/README.md
+++ b/README.md
@@ -174,8 +174,9 @@ this will evaluate to:
 
 You may also use if/then constructs in scope expressions. In this case, the `if`
 field should be a parameter and the `then` must be a scope expression that will
-be subsituted in for the if/then block if the `if` parameter exists and is truthy.
-Given the follwoing example:
+be subsituted in for the if/then block if the `if` parameter is a boolean and is true.
+This will not do any truthiness conversions, you must do that yourself if desired.
+Given the following example:
 
 ```js
   scopes: {AnyOf: ['foo:bar', {if: 'public', then: {AnyOf: []}}]}
--- a/README.md
+++ b/README.md
@@ -168,6 +168,18 @@ this will evaluate to:
   scopes: {AllOf: ['queue:route:foo', 'queue:route:bar']}
 ```
 
+You may also use if/then constructs in scope expressions. In this case, the `if`
+field should be a parameter and the `then` must be a scope expression that will
+be subsituted in for the if/then block if the `if` parameter exists and is truthy.
+Given the follwoing example:
+
+```js
+  scopes: {AnyOf: ['foo:bar', {if: 'public', then: {AnyOf: []}}]}
+```
+
+We can call `authorize({public: true})` and the method call will be permitted
+even if the client does not have the scope `foo:bar`.
+
 In addition, parameters can be nested objects and accessed with dotted notation.
 Given parameters `task` that is `{extra: {treeherder: {symbol: 'F'}}}`, we can do
 the following:
--- a/README.md
+++ b/README.md
@@ -276,7 +276,15 @@ allowed time.
 
 The `req.authorize(params, options)` method returns `true` if the client satisfies
 the endpoint's scope expression. If the client does not satisfy the expression, it
-returns `false` and sends an error message unless `options.noReply` is true.
+throws an error with the code 'AuthorizationError'. You can catch this if you wish
+or let it bubble up and taskcluster-lib-api will return a detailed error message
+to the client.
+
+The AuthorizationError has 3 extra fields to help inspect the results.
+
+**err.scopes:** The scopeset containing the scopes the client has
+**err.expression:** The scope expression that was not satisfied
+**err.missing:** The reduced subset of the expression containing only scopes that were missing
 
 The first argument to `req.authorize` must be an object where the keys are parameters
 to the scope expression defined in the method definition. If any parameters are missing
--- a/README.md
+++ b/README.md
@@ -179,10 +179,10 @@ This will not do any truthiness conversions, you must do that yourself if desire
 Given the following example:
 
 ```js
-  scopes: {AnyOf: ['foo:bar', {if: 'public', then: {AnyOf: []}}]}
+  scopes: {AllOf: [{if: 'private', then: 'foo:bar'}]}
 ```
 
-We can call `authorize({public: true})` and the method call will be permitted
+We can call `authorize({private: false})` and the method call will be permitted
 even if the client does not have the scope `foo:bar`.
 
 In addition, parameters can be nested objects and accessed with dotted notation.
--- a/README.md
+++ b/README.md
@@ -94,9 +94,7 @@ To declare an API method, call `api.declare(options, handler)` with the followin
  * `route` (required) - the URL pattern, with parameters, e.g., `'/object/:id/action/:param'`
  * `params` - patterns for URL parameters (see below)
  * `query` - patterns for query parameters (see below)
- * `scopes` - scopes required for this API endpoint, in disjunctive normal form (see below)
- * `deferAuth` - if true, authentication will not be checked automatically before the handler
-   is invoked.  In this case, the handler must call `req.satisfies()`; see below.
+ * `scopes` - scopes required for this API endpoint, in 'scope expression' form (see below)
  * `stability` - API stability level, defaulting to experimental (see below)
  * `input` - the schema against which the input payload will be validated
  * `skipInputValidation` - if true, don't do input validation (but include the schema in documentation)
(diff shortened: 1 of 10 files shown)


src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
+          var subexpressions = [];
+          template[key].forEach(scope => {
             if (typeof scope === 'string') {
-              return scope.replace(/<([^>]+)>/g, function(match, param) {
-                var value = params[param];
-                return value === undefined ? match : value;
+              subexpressions.push(_splatParams(scope));
+            } else if (_.isObject(scope) && scope.for && scope.in && scope.each) {
+              let subs = _.at(params, scope.in)[0];
+              if (!subs) {
```
djmitche: `|| !_.isArray(subs)` (actually, probably better to give a useful error message here instead of considering it missing)

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
-        if (!(scopesets instanceof Array)) {
-          var params = scopesets;
-          scopesets = _.cloneDeepWith(entry.scopes, function(scope) {
+        req.hasAuthed = false;
+        let missing = [];
+        let hasConditional = false;
+
+        var _splatParams = (scope) => scope.replace(/<([^>]+)>/g, function(match, param) {
+          var value = _.at(params, param)[0];
+          if (value !== undefined) {
```
djmitche: Should probably check for a non-string here and give a useful error message.

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
+              subexpressions.push(_splatParams(scope));
+            } else if (_.isObject(scope) && scope.for && scope.in && scope.each) {
+              let subs = _.at(params, scope.in)[0];
+              if (!subs) {
+                missing.push(scope.in);
+                subexpressions.push(scope);
+                return;
+              }
+              subs.forEach(param => {
+                subexpressions.push(_splatParams(scope.each.replace(`<${scope.for}>`, param)));
```
djmitche: So, answering my earlier question, nesting is not allowed.  That's fine, but should be documented.

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
+              subexpressions.push(_expandExpressionTemplate(scope));
             }
           });
+          return {[key]: subexpressions};
+        };
+
+        var scopeExpression = _expandExpressionTemplate(entry.scopes);
+
+        if (hasConditional && allowLater) {
+          debug('Conditional provided in scopes. Deferring authorization.');
```
djmitche: Why is this the case?  If the value for the conditional wasn't missing, isn't it fine?

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
+          let missing = scopes.removeGivenScopes(result.scopes, scopeExpression);
           res.reportError('InsufficientScopes', [
-            'You do not have sufficient scopes. This request requires you',
-            'to have one of the following sets of scopes:',
-            '{{scopesets}}',
+            'You do not have sufficient scopes. You are missing the following scopes:',
+            '',
+            '{{missing}}',
             '',
             'You only have the scopes:',
```
djmitche: Let's remove `only` here.

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
           return false;
         }
 
-        // If we're not given an array, we assume it's a set of parameters that
-        // must be used to parameterize the original scopesets
-        if (!(scopesets instanceof Array)) {
-          var params = scopesets;
-          scopesets = _.cloneDeepWith(entry.scopes, function(scope) {
+        req.hasAuthed = false;
```
djmitche: I feel like this should be set earlier -- although I guess undefined is falsey too so it's not too important.

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
+            'This request requires you to satisfy this scope expression:',
+            '',
+            '{{scopeExpression}}',
+          ]
+            .join('\n'),  {missing, scopeExpression, scopes: result.scopes});
         }
         return retval;
       };
 
+      req.hasAuthed = true; // No need to check auth unless there are scopes
```
djmitche: It took me a while to figure out that this applies only when `!entry.scopes`, since `req.authorize` will set it back to false.  It also means that for the briefest of moments, `req.hasAuthed` is true even when the req has not been authed.  Maybe expand the conditional below a little?

```
req.hasAuthed = false;

if (!entry.scopes) {
  req.hasAuthed = true;  // No need to check auth if there are no scopes
  next();
} else if (await req.authorize(req.params, {allowLater: true})) {
  next();
}
```

Also, I'm confused how this function appears to only call `next()` in some circumstances.. but that's not new with this PR.

src/api.js
```diff
@@ -529,6 +624,14 @@ var handle = function(handler, context, name) {
   return function(req, res) {
     Promise.resolve(null).then(function() {
       return handler.call(context, req, res);
+    }).then(function() {
+      if (!req.hasAuthed) {
+        // Note: This will not fail the request since a response has already
+        // been sent at this point. It will report to sentry however!
+        // This is only to catch the case where people do not use res.reply()
+        let err = new Error('Deferred auth was never checked and res.reply() not used!');
```
djmitche: `"req.authorized was never called, or some parameters were missing from the request"`

src/api.js
```diff
@@ -683,8 +786,18 @@ API.prototype.declare = function(options, handler) {
...
     }
   });
+  assert(!options.deferAuth,
+    'deferAuth is deprecated! https://github.com/taskcluster/taskcluster-lib-api#request-handlers');
   if ('scopes' in options) {
-    scopes.validateScopeSets(options.scopes);
+    assert(scopes.validExpression(_.cloneDeepWith(options.scopes, scope => {
+      // This makes these template constructs valid parts of an expression
+      if (_.isObject(scope) && scope.for && scope.in && scope.each) {
+        return 'looping-template-construct';
```
djmitche: Perhaps better `['looping', 'template', 'construct']` since it will get turned into an array?

README.md
```diff
@@ -130,29 +128,80 @@ Examples:
...
+  scopes: {AllOf: ['queue:route:foo', 'queue:route:bar']}
+```
+
+You may also use if/then constructs in scope expressions. In this case, the `if`
+field should be a parameter and the `then` must be a scope expression that will
+be subsituted in for the if/then block if the `if` parameter exists and is truthy.
+Given the follwoing example:
+
+```js
+  scopes: {AnyOf: ['foo:bar', {if: 'public', then: {AnyOf: []}}]}
```
jonasfj: I would change this example to be:
```js
scopes: {AllOf: [{if: 'not_public', then: 'foo:bar'}]}
```

README.md
```diff
@@ -130,29 +128,80 @@ Examples:
...
+```
+
+We can call `authorize({public: true})` and the method call will be permitted
+even if the client does not have the scope `foo:bar`.
+
+In addition, parameters can be nested objects and accessed with dotted notation.
+Given parameters `task` that is `{extra: {treeherder: {symbol: 'F'}}}`, we can do
```
jonasfj: Is this smart? 

My idea was that the parameters should be: `{'task.extra.treeherder.symbol': 'F'}` instead....

Note: I'm 100% unsure of all that we're doing here...

README.md
```diff
@@ -220,30 +269,15 @@ includes some additional security token, its duration should be limited to this
...
-If `deferAuth` is set to `true`, then authentication will be postponed to the
-first invocation of `req.satisfies`.  Note that `deferAuth` will not perform
-authorization unless, `req.satisfies({})` is called either without arguments or
-with an object as first argument.  If `deferAuth` is false, then req.params
-will be used as the scope parameters.
-
-Where authorization depends on parameters or the contents of the request, the
-`req.satisfies` method is generally used with `deferAuth: true` in a construct
-like this:
+The `req.authorize(params, options)` method returns `true` if the client satisfies
```
jonasfj: I propose we change this semantics...

Instead we have `req.authorize(params)`, and if authorization fails, then it throw a special error. We can then use `err instanceof AuthorizationError` to catch it in the place where `taskcluster-lib-api` calls the handler..
So if handler doesn't handle the error then it'll propagate up to tc-lib-api will which know how to catch and handle this special error class and knows how to print a nice error message...

This way we get rid of all the `if (!this.authorize(...) { return; }` statements... and it just becomes `this.authorize(...)`.
Handler implementors that wants to do something special can catch the error... and if necessary something like `try { this.authorize(...); ... } finally { ... }` will work nicely...

src/api.js
```diff
@@ -464,56 +511,104 @@ var remoteAuthentication = function(options, entry) {
...
+        var _splatParams = (scope) => scope.replace(/<([^>]+)>/g, function(match, param) {
+          var value = _.at(params, param)[0];
+          if (value !== undefined) {
+            return value;
+          }
+          missing.push(match); // If any are left undefined, we can't be done yet
+          return match;
+        });
+
+        var _expandExpressionTemplate = (template) => {
```
jonasfj: Yeah, let's do this in another file... that way we can have specific unit tests for it too...

jonasfj:
Crazy idea, not sure it's productive at all: use a json-e subset.
The only downside is that `each(...)` expressions are slow to translate because of the regex'es involved... But we can transform those away with a few simple hacks:

```js
// transforms jsone subset to similar structure that's faster to render
// noteably {$map: ..., each(...): ...} is transformed: {$map: ..., each: ..., arg: ...}
let _transform = (template) => {
  if (typeof template === 'string') {
    return template;
  }

  if (template instanceof Array) {
    return template.map(transform);
  }

  if (template.$if) {
    return {
      $if: template.$if,
      then: transform(template.then),
      else: transform(template.else),
    };
  }

  if (!template.$map) {
    const eachKey = Object.keys(template).filter(k => !== '$map')[0];
    const [, arg] = /^each\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(eachKey);
    return {
      $map: template.$map,
      each: template[eachKey],
      arg,
    }
  }
};

// render jsone-like structure from transform given parameters
// this assumes no bugs... error handling would have to be added, but shouldn't
// make this significantly slower.
let _render = (template, params) => {
  if (typeof template === 'string') {
    return template.replace(/\${([a-zA-Z][a-zA-Z0-9]*)}/g, (orig, arg) => {
      return params[arg] || orig;
    });
  }
  
  if (template instanceof Array) {
    return template.map(val => render(val, params)).filter(val => !!val);
  }

  if (template.$if) {
    if (params[template.$if]) {
      return render(template.then, params)
    }
    return render(template.else, params)
  }

  if (template.$map) {
    params[template.$map].map(val =>
      render(template.each, Object.assign({}, params, {[template.arg]: val}))
    );
  }

  let result = {};
  Object.keys(template).forEach(key => {
    result[key] = render(template[key], params);
  });

  return result;
};

// TODO: write a validate() function that ensures we only use the allowed
//       json-e subset, which is:
//       - strings containing ${parameter} (no expressions)
//       - conditinoals {$if: 'parameter', then: ..., else: ...}
//       - maps {$map: 'parameter', each(...): ...}
//       notice that expressions are not allowed in any of these.

// creates a renderer function that given parameters returns the expression.
let createRender = (template) => {
  template = _transform(template); // notice that this only happens at start-up
  return (params) => _render(template, params);
};
```

I'll reiterate I'm not 100% sure about all we're doing here... the conditionals and each constructs, whether we use json-e subset or not, are hard to document... are we better of limiting to this the absolute minimum and simply allow `this.authorize(null, scopeExpression)` so that advanced things can inject a custom scope expression and just document the behavior in the description?

djmitche:
> Crazy idea, not sure it's productive at all: use a json-e subset.

I knew this day would come

README.md
djmitche: I think what you're saying here is, let's don't use `_.at` and instead require all parameters to be named -- but allowing dots in parameters if the user prefers.  I can support that.  `req.params` will never have structure -- it's just strings.  And anything else is provided by the API implementation.

README.md
```diff
@@ -130,29 +128,80 @@ Examples:
...
 
 If scope validation fails, the user is presented with an extensive error
 message indicating the available and required scopes.
 
+Arrays of required scopes can also be templated in the scopes section. When
+this form is used, the parameter must be an array.
+
+```js
+  scopes: {AllOf: [
+    {for: 'route', in: 'routes', each: 'queue:route:<route>'}
```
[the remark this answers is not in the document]
djmitche: The advantage of the way we're introducing this is, we can wait until we find such a case, then add that support to tc-lib-api :)

README.md
djmitche: Surely there's room for a wee `!` operator in there?

djmitche:
Seriously, though, I'm pretty happy with this as-is.  The door is open to allow nesting later, but prohibit it now.  We can optimize the implementation later if necessary, but this seems reasonably fast as-is.  This is sufficiently expressive for all of our uses (including sometimes-public artifacts), at least as far as we've figured out.  And it still remains semantically meaningful, so we can generate readable documentation from it.  If there are a few corner cases that we need to defer to manual overrides and some documentation, I think that's OK (for example, `public` or `not_public`).

I like the *model* of thinking of this as a transformation: given template and parameters, it produces a scope expression, and then that scope expression is validated against the available scopes.  I think the code could better represent those two steps as distinct, unit-testable functions, and perhaps the documentation could define those two stages, as well.  That will be even more important when we move the checking to the Auth service, as it will get the scope expression.

README.md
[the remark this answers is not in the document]
imbstack: I will update documentation to say that this is not recursive and we can make it recursive later if we need it.

p.s. I really hope we don't need it 😰

README.md
imbstack: My thinking with `_.at` is to allow us to just pass the task in directly to `authorize` and grab `task.scopes` or `task.routes` out of it directly. This seems like a common use-case?

djmitche: It doesn't seem to make things too much more complex.  Probably good to point to the `_.at` docs to indicate that it can also index array.

src/api.js
imbstack: The general issue here is that if the conditional exists, we don't know that we are allowed to ignore results from the initial authorize that happens automatically. With parameters, we know that we can skip authorizing if any are missing, but with `if/then` we can't really know. 

An example is `scopes: {AnyOf: ['whatever', {if: 'public', then: {AllOf: []}}]}`. When we initially call the function, we don't know `public` yet, so this evaluates to `{AnyOf: ['whatever']}` which will fail for public requests with no scopes. We will never get to the part where we call `authorize` later in this case.

Now, obviously we can fix this specific case by doing jonas' more clever  `{AllOf: [{if: 'not_public', then: 'foo:bar'}]}`, but now we're at the point where we're requiring people to craft scope expressions in such a way that they must always succeed when initially evaluated  by the endpoint in order to allow them to call `authorize` explicitly later. Really what we could use here is adding back in a `deferAuth` flag I guess? I'm sorta not loving the `if/then` construct we have. It complicates things a good deal.

imbstack: We figured this out in irc!

src/api.js
imbstack: It doesn't actually get turned into an array unfortunately. It merges its resulting array into the array it is a part of.

README.md
```diff
@@ -94,9 +94,7 @@ To declare an API method, call `api.declare(options, handler)` with the followin
  * `route` (required) - the URL pattern, with parameters, e.g., `'/object/:id/action/:param'`
  * `params` - patterns for URL parameters (see below)
  * `query` - patterns for query parameters (see below)
- * `scopes` - scopes required for this API endpoint, in disjunctive normal form (see below)
- * `deferAuth` - if true, authentication will not be checked automatically before the handler
-   is invoked.  In this case, the handler must call `req.satisfies()`; see below.
+ * `scopes` - scopes required for this API endpoint, in 'scope expression' form (see below)
```
jonasfj: on vidyo we discussed adding a `terms` property that makes from parameters to explanations. This could be done in a follow up..

imbstack:
I've updated this with changes from our conversations. Some decently large changes have landed, so please review the new commits carefully. We should squash these commits into one before landing!

src/api.js
```diff
@@ -394,126 +432,152 @@ var remoteAuthentication = function(options, entry) {
...
-                   missing.join('", and\n    - "') + '"';
-          })).join('\n'),  {scopesets, scopes: result.scopes});
+            'This request requires you to satisfy this scope expression:',
+            '',
+            '{{expression}}',
+          ].join('\n');
+          throw err;
         }
-        return retval;
+        return authed;
```
djmitche: so basically, `return true` at this point..

src/schemas/api-reference.json
```diff
@@ -106,17 +148,8 @@
...
-                            "type": "array",
-                            "description": "A set of scopes that grants access if all the scopes in the set is satisfied.  Scopes must be composed of printable ASCII characters and spaces.",
-                            "items": {
-                                "type": "string",
-                                "description": "Scope identifier",
-                                "pattern": "^[\\x20-\\x7e]*$"
-                            }
-                        }
+                        "description": "Scope expression template specifying required scopes for a method. Not provided if authentication isn't required.",
+                        "$ref": "#/definitions/scopeExpressionTemplate"
```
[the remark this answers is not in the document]
imbstack: I think the entire service needs to be ported to scope expressions in one fell swoop. Docs is the only consumer of these references (afaik) and my docs changes will handle both sorts of expressions/sets until we finish porting all services.

imbstack:
[here](https://github.com/taskcluster/taskcluster-queue/commit/8cc588f31bc015f9fd41004e964de624a539a76d) is the update we'll make to queue when this lands and [here](https://github.com/taskcluster/taskcluster-index/compare/what-will-happen-with-scopes) is what will happen to index! I think queue is the hardest one we have so we should be ok with every other service. Tests are green in these services with these changes.

README.md
```diff
@@ -220,30 +274,23 @@ includes some additional security token, its duration should be limited to this
...
 
-The `req.satisfies(.., noReply)` method returns `true` if the client satisfies
-one of the scopesets. If the client does not satisfy one of the scopesets, it
-returns `false` and sends an error message unless `noReply` is true. 
-
-The first argument to `req.satisfies` can be a scopeset (in disjunctive normal
-form as described above).  Or, it can be an object, in which case the method
-will assume this object is a mapping from scope parameters (`<name>`) to
-values.
+The `async req.authorize(params, options)` method returns `true` if the client satisfies
```
jonasfj: don't return `true`, I suggest we leave the return-value unspecified... In the future we might return `clientId` and set of scopes used to satisfy the expression... or something like that... as this is probably what taskcluster-auth will return...

README.md
```diff
@@ -220,30 +274,23 @@ includes some additional security token, its duration should be limited to this
...
+**err.missing:** The reduced subset of the expression containing only scopes that were missing
 
-```js
-  if (!req.satisfies({hookGroupId, hookId})) {
-    return;
-  }
-```
+The first argument to `req.authorize` must be an object where the keys are parameters
+to the scope expression defined in the method definition. If any parameters are missing
+when you call this, the check will error unless you pass `options.allowLater`. This is
```
jonasfj: When, if ever, is this needed?

src/expressions.js
```diff
@@ -0,0 +1,45 @@
+const scopes = require('taskcluster-lib-scopes');
+const _ = require('lodash');
+
+export const splatParams = (scope, params, missing) => scope.replace(/<([^>]+)>/g, (match, param) => {
```
jonasfj: doc comment please

src/expressions.js
```diff
@@ -0,0 +1,45 @@
...
+  const value = _.at(params, param)[0];
+  if (value !== undefined) {
+    return value;
+  }
+  missing.push(match); // If any are left undefined, we can't be done yet
+  return match;
+});
+
+export const expandExpressionTemplate = (template, params, missing) => {
```
jonasfj: doc comment please

src/expressions.js
```diff
@@ -0,0 +1,45 @@
...
+  if (typeof template === 'string') {
+    return splatParams(template, params, missing);
+  } else if (_.isObject(template) && template.for && template.in && template.each) {
+    let subs = _.at(params, template.in)[0];
+    if (!subs) {
+      missing.push(template.in);
+      return;
+    }
+    return subs.map(param => splatParams(template.each.replace(`<${template.for}>`, param), params, missing));
```
jonasfj: Are you sure we shouldn't do:
```js
const param2 = Object.assign({}, params);
subs.map(param => {
  param2[template.for] = param;
  return splatParams(template.each, param2);
});
```

It's is more expensive.. but probably not much...

jonasfj: better name than `param2` obivously...

jonasfj: Note: this is strictly more expressive...

src/expressions.js
```diff
@@ -0,0 +1,45 @@
...
+    const conditional = _.at(params, template.if)[0];
+    if (conditional === true) {
+      // Only do this if the conditional exists and is literally true
+      return expandExpressionTemplate(template.then, params, missing);
+    } else if (conditional === undefined) {
+      // In the case that the conditional is false, do nothing. If it is missing, we say so
+      missing.push(template.if);
+    } else if (conditional !== false) {
+      throw new error(`conditional values must be booleans! ${template.if} is a ${typeof template.if}`);
+    }
```
jonasfj: if you intentionally return `undefined` please do so...

src/expressions.js
```diff
@@ -0,0 +1,45 @@
...
+  } else {
```
jonasfj: I would make two cases `else if (template.AnyOf) {` and for `AllOf`...

That way you don't have to call `Object.keys(...)`, it'll likely be faster, and easier to read... even if it's a bit more code...

src/expressions.js
```diff
@@ -0,0 +1,45 @@
...
+    const key = Object.keys(template)[0];
+    let subexpressions = [];
+    template[key].forEach(scope => {
```
jonasfj: instead of `scope` maybe use a variable like `subexpr`

src/expressions.js
```diff
@@ -0,0 +1,45 @@
...
+      const results = expandExpressionTemplate(scope, params, missing);
+      if (results) {
+        subexpressions = subexpressions.concat(results);
```
jonasfj: Isn't it only `concat` if `results` is an array? it could be a string right?

src/schemas/api-reference.json
```diff
@@ -1,9 +1,51 @@
...
     "type": "object",
+    "definitions": {
+      "subExpression": {"type": "array", "items": {"$ref": "#/definitions/scopeExpressionTemplate"}},
+      "scopeExpressionTemplate": {
+        "oneOf": [
+          {"type": "string", "pattern": "^[\\x20-\\x7e]*$"},
+          {
+            "type": "object",
+            "properties": {
+              "AnyOf": {"$ref": "#/definitions/subExpression"}
```
jonasfj: Please copy paste: `{"type": "array", "items": {"$ref": "#/definitions/scopeExpressionTemplate"}}`

Then we don't have to discuss the name `subExpression` :)

src/schemas/api-reference.json
```diff
@@ -1,9 +1,51 @@
...
+            },
+            "required": ["AnyOf"],
+            "additionalProperties": false
+          },
+          {
+            "type": "object",
```
jonasfj: please consider adding some descriptions...

src/expressions.js
imbstack: It turns out that [concat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) does the right thing!

PR #72, opened 2018-01-27
Tweak scope expressions, pre-process, allow if-then-else, allow strin…

jonasfj:
…gs, restrict param names

This grew to also tweak things...

### DIFF
--- a/src/expressions.js
+++ b/src/expressions.js
@@ -1,73 +1,242 @@
-const scopes = require('taskcluster-lib-scopes');
 const _ = require('lodash');
 
-/**
- * Given a string, potentially with parameters included (indicated by <...>) and
- * a set of parameters, return a new string that has replaced the parameters in the string
- * with the value of the parameter in the parameters object. The third argument is a list
- * that will have the names of any missing parameters appended to it.
- *
- * Example:
- *
- * splatParams('abc:<foo>:<bar>', {foo: 'def', bar: 'ghi'}, []) -> 'abc:def:ghi'
- *
- */
-export const splatParams = (scope, params, missing) => scope.replace(/<([^>]+)>/g, (match, param) => {
-  const value = _.at(params, param)[0];
-  if (value !== undefined) {
-    return value;
+const validRootTemplate = (template) =>
+  validateTemplate(template) && (
+    typeof template === 'string' ||
+    template.hasOwnProperty('AllOf') ||
+    template.hasOwnProperty('AnyOf') ||
+    template.hasOwnProperty('if')
+  )
+;
+
+/** Validate a scope expression template */
+const validateTemplate = (template) => {
+  if (typeof template === 'string') {
+    return true;
   }
-  missing.push(match); // If any are left undefined, we can't be done yet
-  return match;
-});
+  const keys = _.keys(template);
+  if (_.xor(keys, ['AllOf']).length === 0) {
+    return template.AllOf.every(validateTemplate);
+  }
+  if (_.xor(keys, ['AnyOf']).length === 0) {
+    return template.AnyOf.every(validateTemplate);
+  }
+  const paramPattern = /^[a-zA-Z][a-zA-Z0-9_]*$/;
+  if (_.xor(keys, ['for', 'in', 'each']).length === 0) {
+    return (
+      paramPattern.test(template.for) &&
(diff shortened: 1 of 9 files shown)


jonasfj:
yeah, we probably have to do a major release...

jonasfj:
note this ended up being a complete rewrite of `expressions.js`, so don't view the diff when reviewing it..

Note: if you feel like tweaking this by all means hope in :)

djmitche:
Why does a "tweak" require a major release?

src/schemas/api-reference.json
```diff
@@ -161,7 +166,12 @@
                     },
                     "scopes": {
                         "description": "Scope expression template specifying required scopes for a method. Not provided if authentication isn't required.",
-                        "$ref": "#/definitions/scopeExpressionTemplate"
+                        "oneOf": [
```
djmitche: Why is this not just scopeExpressionTemplate?  We allow "if" at the top level but not "for"?

djmitche:
I see that this adds typing to params -- that's pretty awesome and will catch some bugs!

It looks like the changes are:
 * Pre-compile scope expressions for efficiency
 * Check types of parameters (string, boolean, array)
 * [BREAKING] limit parameter names, most importantly excluding `.`, thus prohibiting use of dotted paths
 * Allow strings and `{if: .. then: ..}` at the top level of a scope expression (but not `{for: ..}`)

Splitting that up would be hard, but it would be good to highlight the breaking change at least in the commit message.

NOTE: thinking ahead to when we start doing this analysis in tc-auth instead of in the services, it might make sense to have this in tc-lib-scopes?  I suppose we can move the code at that time.

src/schemas/api-reference.json
imbstack: I think `for` is not allowed at top level because it evaluates to an array of strings rather than a scope expression itself.

imbstack: I think jonas is open to changing that however.

imbstack: And maybe the more I think about it I like it more.

src/expressions.js
```diff
@@ -1,73 +1,242 @@
...
+  }
+  throw new Error('extractParams expects a compiled scope expression templates');
+};
+
+/** Render a scope expression from a compiled scope expression template and parameters */
+const render = (compiledTemplate, params) => {
+  const ctmpl = compiledTemplate;
+  if (ctmpl instanceof Array) {
+    return ctmpl.map((value, i) => {
+      if (i % 2 === 1) {
```
imbstack: All of this stuff with modulo and whatever is making me nervous. It seems like there must be a more obvious way to do this kinda stuff?

src/schemas/api-reference.json
djmitche: No, that reason makes sense.  I'd say "add a comment" but of course, it's JSON :)

djmitche:
Also, I think we've iterated on these expressions enough.. let's deploy 'em already.  If we later decide to allow top-level `for`, that'd be an expansion of capabilities..

src/schemas/api-reference.json
jonasfj: So there is another thing we could do, instead of for and if, we make them conditions on AllOf/AnyOf, like:

`{AllOf: [...], if: 'param'}` and 
`{AllOf: '...<var>...', for: 'var', in: 'param'}`

This way we don't have to merge upwards when evaluating... and AnyOf/AllOf are used to indicate if the condition is trivial satisfied or rejected if the array in "for" is empty or the condition in "if" false...

It would make if/for side conditions on the AllOf/AnyOf constructs.... I'm not sure it's more logical. Which is what matters as it's purely a syntactical thing...

Which is also why I stopped thinking about it... As we could also start questioning why AllOf and AnyOf starts with a capital letter...

imbstack: They start with a capital letter only because they did in the spec in rfc-23. I don't mind either way.

src/expressions.js
jonasfj: Every second string is a parameter... I actually think it's pretty sane..

Yes, the code is unreadable, but if this is wrong tests will definitely catch it :)

jonasfj: +1, if you find a better way to do this... I do care about speed of the `render` function...

PR #74, opened 2018-02-04
I made a bug in the changes when I tweaked things. I totally got the parameter substitution wrong...

jonasfj:
On the plus side we now only have the `details` that we explicitly want to send
as parameters... Before we would get `message` and `code` this way too.

Anyways, this should be a non-controversial fix.


src/api.js
```diff
@@ -373,7 +373,7 @@ var createRemoteSignatureValidator = function(options) {
  * client satisfies the scope expression in `options.scopes` after the
  * parameters denoted by `<...>` and `{for: ..., each: ..., in: ...}` are
  * substituted in. If the client does not satisfy the scope expression, it
- * throws an expressions.AuthorizationError.
+ * throws an Error with code = 'AuthorizationError'.
```
imbstack: Oh, nice catch.

PR #77, opened 2018-02-06
Bug 1436160: don't continue on if pattern is invalid

djmitche:


src/api.js
```diff
@@ -195,8 +195,7 @@ let queryValidator = function(options = {}) {
         }
         // Unsupported option
         errors.push('Query-string parameter: ' + key + ' is not supported!');
```
jonasfj: Add an early `return;` instead...

djmitche:
Test failures are because this isn't in the upstream repo.

PR #78, opened 2018-02-20
Update auth tests to be easier to understand

imbstack:
I'm working on making the tests easier to understand as a baseline before I add more tests to this. Does this change make it easier to understand to you? For me the improvement is that each handler is tied closely to the tests of it rather than running all over this file trying to find the handler for a test and vice-versa. The downside is that this adds more indentation.


test/auth_test.js
```diff
@@ -276,677 +61,837 @@ suite('api/auth', function() {
...
+    route:  '/test-deprecated-satisfies',
+    handler: (req, res) => {
+      if (req.satisfies([])) {
+        res.status(200).json({ok: true});
+      }
+    },
+    tests: [
+      {
+        label: 'function that still uses satisfies fails',
+        testfactory: url => () => {
```
jonasfj: I think you could make this declarative too...

As in list what should be I've to hawk.. what should be expected as result... Status code etc...

And then just have a function as fallback...

jonasfj: Also a different name than test factory please...

djmitche:
QED :)

PR #80, opened 2018-02-22
fixes Bug 1437193

Aksh77:


djmitche:
I did a little experiment:
```javascript
const express = require('express');
const app = express();
app.get('/', (req, res) => { 
  res.status(400);
  res.status(200);
  res.send('Hello World!');
});
app.listen(3000, () => console.log('Example app listening on port 3000!'));
```

and this does not cause an error.  But changing that first `res.status` line to
```javascript
  res.status(400).json({bad: 1});
```
does cause an exception to be raised.  So it seems that just calling `res.status(..)` doesn't actually send anything, and calling it multiple times with different values isn't an error.

Looking back at the original bug, the problem was in calling `res.reportError`, which eventually calls `res.status(..).json(..)`:
https://github.com/taskcluster/taskcluster-lib-api/blob/058a6fe8e34045978cbc31954eb57465320b068a/src/errors.js#L72-L74
after which the taskcluster-notify service sent the email and then [called res.sendStatus(200)](
https://github.com/taskcluster/taskcluster-notify/blob/269c027d6f2394900a2d4ffb7ed84ee792564a01/src/api.js#L36)

The [express docs](http://expressjs.com/en/api.html#res.sendStatus) define that as
```javascript
res.sendStatus(200); // equivalent to res.status(200).send('OK')
```
So I think that would reduce to
```javascript
const express = require('express');
const app = express();
app.get('/', (req, res) => {
  res.status(400).json({error: 'yep'});
  res.status(200).send('OK');
});
app.listen(3000, () => console.log('Example app listening on port 3000!'));
```
but as I noted above, that does give me an error:
```
Error: Can't set headers after they are sent.
```

i returned to taskcluster-lib-notify, running it locally with:
```javascript
api.declare({
  method:       'post',
  route:        '/email',
  name:         'email',
  scopes:       'notify:email:<address>',
  title:        'Send an Email',
  description: [
    'Send an email to `address`. The content is markdown and will be rendered',
    'to HTML, but both the HTML and raw markdown text will be sent in the',
    'email. If a link is included, it will be rendered to a nice button in the',
    'HTML version of the email',
  ].join('\n'),
}, async function(req, res) {
  debug(`Received request to send email to ${req.body.address}`);
  res.reportError('InputError', 'uhoh', {});
  res.sendStatus(200);
});
```
which should do about the same thing.  Yet, I get no error in the logs when I run this (and I get the 400 InputError HTTP response).  It's the same version of express in both cases (4.16.2).  Even changing that around to
```javascript
  debug(`Received request to send email to ${req.body.address}`);
  //res.reportError('InputError', 'uhoh', {});
  res.status(400).json('bad req');
  res.status(200).send('OK');
```
doesn't produce an error.  @Aksh77 - I think this is the thing to understand before we can solve this bug... can you pick it up from here?

Aksh77:
Thanks @djmitche ! I will look into it and try to correct it.

PR #82, opened 2018-02-26
Update api.js

guptatrisha97:
Added code to prohibit defining API method twice

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -782,7 +782,7 @@ API.prototype.declare = function(options, handler) {
     options.output = this._options.schemaPrefix + options.output;
   }
   this._entries.push(options);
-  var valueArr = options.map((item) => {return item.name});
+  var valueArr = options.map((item) => {return item.name};);
   var isDuplicate = valueArr.some(function(item, idx) { 
     return valueArr.indexOf(item) != idx; 
   });
--- a/src/api.js
+++ b/src/api.js
@@ -790,7 +790,6 @@ API.prototype.declare = function(options, handler) {
     throw new Error('This function has already been declared');
   }
 };
-};
 
 /**
  * Construct a router that can be mounted on an express application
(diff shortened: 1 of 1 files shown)


src/api.js
```diff
@@ -782,6 +782,13 @@ API.prototype.declare = function(options, handler) {
     options.output = this._options.schemaPrefix + options.output;
   }
   this._entries.push(options);
+  var valueArr = options.map((item) => {return item.name;});
```
djmitche: `options` here is an object with properties -- there's no need to use `map` to find its name attribute.  Just write `options.name`.

src/api.js
```diff
@@ -782,6 +782,13 @@ API.prototype.declare = function(options, handler) {
...
+  var isDuplicate = valueArr.some(function(item, idx) {
```
djmitche: This seems to be searching for a duplicate entry in `valueArr`, which is an array of .. well, I'm not sure what map returns when run over an object.  I think what you want here is `this._entries`, which has the list of entries that have already been defined.

src/api.js
```diff
@@ -782,6 +782,13 @@ API.prototype.declare = function(options, handler) {
     options.output = this._options.schemaPrefix + options.output;
   }
   this._entries.push(options);
+  var valueArr = options.map((item) => {return item.name;});
+  var isDuplicate = valueArr.some(function(item, idx) { 
+    return valueArr.indexOf(item) != idx;
```
djmitche: You will want to see if there is an item in `this._entries` which has the name `options.name`.  You'll need to do that *before* `options` is pushed onto `this._entries`, though.

src/api.js
guptatrisha97: in that case, where will I check for options.name? Would it be var isDuplicate = this._entries.some(function(options.name, idx) ?

djmitche: You'll check that below, as I suggested.

djmitche:
Picked up by @kritisingh1 in #95

PR #83, opened 2018-03-03
add header

fiennyangeln:
Bug 1442636
Add header
@djmitche please review :)

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -848,6 +848,7 @@ API.prototype.router = function(options) {
   if (options.allowedCORSOrigin) {
     router.use(function(req, res, next) {
       res.header('Access-Control-Allow-Origin',   options.allowedCORSOrigin);
+      res.header('Access-Control-Max-Age', 900);
       res.header('Access-Control-Allow-Methods', [
         'OPTIONS',
         'GET',


djmitche:
@fiennyangeln this looks good, but can you modify the [commit message](https://docs.taskcluster.net/manual/design/devel/best-practices/commits)?  Then I'll merge it and make a release, and you can start working on upgrading services :)

fiennyangeln:
@djmitche Ok modified based on the commit message guidelines. Moreover, does it have to be like that for other commit ? :) And, which services do you mean? :D

djmitche:
That updated commit message is great.  Yes, ideally all commits have a good title and description.

The services are in the first three categories in https://docs.taskcluster.net/reference (Taskcluster Platform, Core Services, and Integrations).  Plus ec2-manager, which doesn't show up there for reasons we won't get into.

fiennyangeln:
@djmitche do you mean adding the value to the option to be passed when creating API on that service? :D Because it seems like it is already defaulted to use the allowedCORSOrigin and on the Hooks, auth, queue repo, they dont send the value, so the default one (the allowed cors origin) is used. Is my understanding correct ? :D https://github.com/taskcluster/taskcluster-lib-api/blob/master/src/api.js#L811

PR #84, opened 2018-03-04
verify that same context is passed

kritisingh1:
Bug 1440800
Matches the `context` attribute of the API declared to that of its setup.
@djmitche Please review and let me know further changes.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -810,7 +810,7 @@ API.prototype.router = function(options) {
   options = _.defaults({}, options, {
     inputLimit:           '10mb',
     allowedCORSOrigin:    '*',
-    context:              {},
+    context:              this._options.context,
     nonceManager:         nonceManager(),
     signatureValidator:   createRemoteSignatureValidator({
       authBaseUrl:        options.authBaseUrl || AUTH_BASE_URL,
--- a/src/api.js
+++ b/src/api.js
@@ -823,10 +823,10 @@ API.prototype.router = function(options) {
       'Context must have declared property: \'' + property + '\'');
   });
 
-  Object.keys(options.context).forEach(function(property) {
+  Object.keys(options.context).forEach(property => {
     assert(this._options.context.indexOf(property) !== -1,
-      'Context has unexpected property: \'' + property + '\'');
-  }.bind(this));
+      `Context has unexpected property: ${property}`);
+  });
 
   // Create caching authentication strategy if possible
   if (options.clientLoader || options.credentials) {
--- a/src/api.js
+++ b/src/api.js
@@ -810,7 +810,7 @@ API.prototype.router = function(options) {
   options = _.defaults({}, options, {
     inputLimit:           '10mb',
     allowedCORSOrigin:    '*',
-    context:              this._options.context,
+    context:              {},
     nonceManager:         nonceManager(),
     signatureValidator:   createRemoteSignatureValidator({
       authBaseUrl:        options.authBaseUrl || AUTH_BASE_URL,
@@ -822,6 +822,11 @@ API.prototype.router = function(options) {
     assert(options.context[property] !== undefined,
       'Context must have declared property: \'' + property + '\'');
   });
+
+  Object.keys(options.context).forEach(function(property) {
+    assert(this._options.context.indexOf(property) !== -1,
+      'Context has unexpected property: \'' + property + '\'');
+  }.bind(this));
 
   // Create caching authentication strategy if possible
   if (options.clientLoader || options.credentials) {
--- a/test/context_test.js
+++ b/test/context_test.js
@@ -137,7 +137,7 @@ suite('API (context)', function() {
         },
       });
     } catch (err) {
-      if (/Context has unexpected property: 'prop3'/.test(err)) {
+      if (/Context has unexpected property: prop3/.test(err)) {
         return; //expected error
       }
     }
--- a/test/context_test.js
+++ b/test/context_test.js
@@ -84,7 +84,9 @@ suite('API (context)', function() {
         },
       });
     } catch (err) {
-      return; // expected error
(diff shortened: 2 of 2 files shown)


src/api.js
```diff
@@ -809,7 +809,7 @@ API.prototype.router = function(options) {
   options = _.defaults({}, options, {
     inputLimit:           '10mb',
     allowedCORSOrigin:    '*',
-    context:              {},
```
djmitche: The `context` argument to `API()` is an array of strings (names of expected context objects).  The `context` argument to `api.router()` is an object, with context values at each property.

```javascript
const api = API({
  context: ['database', 'cache'],
});

api.router({
  context: {
    'database': new Database(),
    'cache': new Cache(),
  },
});
```

So making one (the object) default to the other (the array) doesn't make much sense.

kritisingh1:
@djmitche I have added the test for checking unknown entry passed to context in `api.router`. Please review and suggest further changes. Thanks :)

test/context_test.js
```diff
@@ -108,4 +108,30 @@ suite('API (context)', function() {
...
+        validator: validate,
+        context: {
+          prop3: 'value3',
+        },
+      });
+    } catch (err) {
+      return; //expected error
+    }
+    assert(false, 'Expected an error!');
+  });
```
djmitche: This is a good new test.  However, it's not catching the error you want it to catch!  If you modify the `catch` section like this:
```
    } catch (err) {
      if (/Context has unexpected property: 'prop3'/.test(err)) {
        return; //expected error
      }   
    }   
```

then the test will fail.  It would be helpful to also make a similar modification (but, looking for a different message) to the "Context properties can be required" test that you copied this one from.

kritisingh1:
@djmitche I have made the changes as suggested. Please review. Thanks!

djmitche:
Did you run the tests?  I get..
```

  1) API (context)
       Context properties can be required:

      AssertionError: Expected an error!
      + expected - actual

      -false
      +true
      
      at Suite._callee2$ (test/context_test.js:91:5)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:65:40)
      at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:299:22)
      at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:117:21)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13

  2) API (context)
       Context entry should be known:

      AssertionError: Expected an error!
      + expected - actual

      -false
      +true
      
      at Suite._callee4$ (test/context_test.js:143:5)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:65:40)
      at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:299:22)
      at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:117:21)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13

```

kritisingh1:
@djmitche The expected behavior is to make those two tests fail, right? If I pass `false` in assert, the tests fail and if I pass `true` in assert, those pass. So what should be done ideally.. please guide me. Thank you!

djmitche:
Nope, the tests were right with `assert(false, ..)`.  We expect to get an error and return a few lines earlier, so if we get to the `assert(false, ..)` then we haven't gotten an error.

The remaining work is to adjust the code so that the tests pass.

kritisingh1:
@djmitche I had made the required changes some days ago. Please review. Thanks :)

test/context_test.js
```diff
@@ -111,4 +114,34 @@ suite('API (context)', function() {
...
+        validator: validate,
+        context: {
+          prop3: 'value3',
+        },
+      });
+    } catch (err) {
+      if (/Context has unexpected property: 'prop3'/.test(err)) {
+        return; //expected error
+      }
+      return;
```
djmitche: This makes the test return success even if the expected error didn't occur.  With this return removed(and the one in the previous test, too), the test doesn't pass.  That's a hint that the solution in `src/api.js` isn't correct, as I suggested 7 days ago. I double-checked this by looking for the string "Context has unexpected property".  It only occurs once in the source code (right here).  So nothing could ever throw that error!

src/api.js
```diff
@@ -823,6 +823,11 @@ API.prototype.router = function(options) {
       'Context must have declared property: \'' + property + '\'');
   });
 
+  Object.keys(options.context).forEach(function(property) {
+    assert(this._options.context.indexOf(property) !== -1,
+      'Context has unexpected property: \'' + property + '\'');
```
djmitche: You can use a string template for this: `Context has unexpected property '${property}'`.

src/api.js
```diff
@@ -823,6 +823,11 @@ API.prototype.router = function(options) {
...
+  }.bind(this));
```
djmitche: If you use an arrow function (`property => { .. }`) instead of the function keyword (`function(property) { .. }`), there is no need to bind.

kritisingh1:
@djmitche I have updated the PR. Please have a look. Sorry for delay. Thank you for patience :)

djmitche:
Did you forget to push? :)

kritisingh1:
No, actually there were some internet issues so the comment was updated a bit late. Sorry. :)

djmitche:
That will be version 8.0.0 of this library.  Do you want to try upgrading, say, taskcluster-secrets or taskcluster-hooks to use this version?  The patch to upgrade will be simple, but we'll need to make sure that the tests all pass once that's done.

PR #85, opened 2018-03-06
Add a "name" to the api for references

imbstack:
This will allow the clients to figure out the url to access without inferring it from the baseUrl we currently provide. I also get the feeling that this could be used to reduce the amount of arguments we need to pass to api at instantiation and definition in general if someone took a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ let API = require('taskcluster-lib-api');
 let api = new API({
   // Title and description for docs
   title: 'My API',
+  name: 'my-api', // Must match /^[a-zA-Z][a-zA-Z0-9_-]*$/
   description: [
     "Long string with **markdown** support, used for writing docs",
     "typically written using [].join('\n') to allow for long strings"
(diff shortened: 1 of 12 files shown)


imbstack:
Now that I look into integrating this into clients more, I think this is incomplete. It probably needs either one of:

1. A `version` field as well that we normally set to `v1` that gets appended to the url e.g. `taskcluster.net/<service_name>/<version>`
2. Force `name` to include a single slash where the version is after the slash
3. Get rid of versions in our apis or something? (this one seems like the hardest choice but I know we've talked about doing this before.

djmitche:
How will this work in clients?  I'm missing the larger plan.

Regarding version, we basically hardcode that to `v1` everywhere, so a little more hard-coding is no problem in my view.

imbstack:
Clients will now allow setting a `apiBase` or whatever (we still haven't settled on a good name). That will be used in combination with this value to construct a baseUrl for each service:

#### client.js

```js
const baseUrl = `${this._config.apiBase}/${reference.name}/${reference.version}`;
// This would construct something along the lines of:
// https://taskcluster.net/auth/v1
```

Currently we can pull these values out of the `baseUrl` we set for each reference, but it would be nicer if we just set them explicitly as well.

djmitche:
OK, thanks.  In that case, option 1 (a version field) makes sense to me.

IIRC the clients are going to temporarily special-case `apiBase === "taskcluster.net"` to generate the existing, hostname-based URLs, right?

jonasfj:
Presumably many things will change if we change the version... probably the reference format would need to be bumped.

So why not just say that `reference.version == 0` implies:
```js
const baseUrl = `${this._config.apiBase}/${reference.name}/v1`;
```

Note: schema for references says `version: 0` is required. We never made it `1` because it's not really super stable yet. At least not as far as the public is aware.

imbstack:
For the time being I think they will just use `baseUrl`, yeah.

PR #89, opened 2018-03-29
Fail if same route and method is declared twice

fiennyangeln:
Done by modifying the declare method in API, add test, and fix
test that uses same method and route declaration

Bug 1449697


fiennyangeln:
@djmitche Please review:D

fiennyangeln:
I think the publish test fail because we declare the route test4/ and test3/ twice. Would it be ok if i change the name as well?

PR #90, opened 2018-03-30
Bug 1443017 :Use a mock AWS library to test publishing API definitions

akshitac8:
This PR is in reference to the issue [bug 1443017 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1443017).
It still needs some changes as it shows timeout at publish.
Kindly review the changes.

### DIFF
--- a/test/publish_test.js
+++ b/test/publish_test.js
@@ -1,7 +1,7 @@
 suite('api/publish', function() {
   var subject         = require('../');
   var config          = require('typed-env-config');
-  var aws             = require('aws-sdk');
+  var awsmock         = require('mock-aws-s3');
   var assert          = require('assert');
   var Promise         = require('promise');
 
@@ -10,7 +10,7 @@ suite('api/publish', function() {
   if (!cfg.aws || !cfg.referenceTestBucket) {
     console.log('Skipping \'publish\', missing config file: ' +
                 'taskcluster-base-test.conf.json');
-    this.pending = true;
+  this.skip = true;
   }
 
   // Test simple method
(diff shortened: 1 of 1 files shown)


test/publish_test.js
```diff
@@ -10,7 +10,7 @@ suite('api/publish', function() {
   if (!cfg.aws || !cfg.referenceTestBucket) {
     console.log('Skipping \'publish\', missing config file: ' +
                 'taskcluster-base-test.conf.json');
-    this.pending = true;
+  this.skip = true;
```
djmitche: `skip` is a method; `pending` is a property.  I don't know why, but that's how mocha works.  Anyway, setting `skip = true` won't do anything.

test/publish_test.js
```diff
@@ -102,20 +102,31 @@ suite('api/publish', function() {
...
+      .then(function() {
+        // Get the file... we don't bother checking the contents this is good
+        // enough
+        var s3 = new aws.S3({accessKeyId: 'fake', secretAccessKey: 'fake'});
+        // return s3.getObject({
+        //  Bucket:     cfg.referenceTestBucket,
+        //   Key:        'base/test/simple-api.json',
+        // }).promise();
+
+        awsmock.mock('S3', 'getObject', function(param, callback) {
```
djmitche: Interesting that you set up the mock *after* making the call it is mocking..

akshitac8:
It is giving me output publish timeout and not passing the test I guess.

djmitche:
Better keep hacking, then :)

PR #92, opened 2018-04-03
Bugfix json subs

djmitche:


djmitche:
Hm, now this will escape the JSON too:

```
\[
  1,
  2,
  3
\]
```

djmitche:
Maybe @eliperelman is right in https://github.com/taskcluster/taskcluster-rfcs/issues/117, and we should just render these as preformatted text, not markdown.

djmitche:
I'm going to land this anyway, as it's ugly but not vulnerable.

PR #94, opened 2018-04-05
update taskcluster-lib-validate to 3.1.0

ydidwania:


djmitche:
Oh, good thinking.  We'll need a new version of lib-api at the same time.

djmitche:
Oh, Travis, Y U so slow..

djmitche:
v8.2.0, should be published in a few minutes

djmitche:
it's out now

ydidwania:
Should I update this along with `taskcliuster-lib-validate` ?

djmitche:
Yes please.  Let's start with just one service and see how it goes.  Queue might be a good choice.

PR #95, opened 2018-04-07
prevent method declaration with duplicate name

kritisingh1:
Reference : BUG 1434583
This would prevent declaration of a method with same name. For now, the tests fail as there are already many test cases with api declarations with same name.
@djmitche Please review and guide me on how to change the tests. Listing in separate files could be a 
solution however not efficient enough. Or else we can rename the duplicate test declarations. After this is fixed I would add the corresponding tests too.
Thank you :D

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -789,13 +789,10 @@ API.prototype.declare = function(options, handler) {
   if (this._entries.filter(entry => entry.route == options.route && entry.method == options.method).length > 0) {
     throw new Error('Identical route and method declaration.');
   }
-  if (this._entries.map(a => a.name).some(function(element) {
-    return element == options.name;
-  })) {
+  if (this._entries.some(entry => entry.name === options.name)) {
     throw new Error('This function has already been declared.');
-  } else {
-    this._entries.push(options);
   }
+  this._entries.push(options);
 };
 
 /**
--- a/src/api.js
+++ b/src/api.js
@@ -789,7 +789,19 @@ API.prototype.declare = function(options, handler) {
   if (this._entries.filter(entry => entry.route == options.route && entry.method == options.method).length > 0) {
     throw new Error('Identical route and method declaration.');
   }
-  this._entries.push(options);
+  var testArray = this._entries.map(a => a.name);
+  var length = testArray.length;
+  var isDuplicate = 0;
+  for (var i = 0; i < length; i++) {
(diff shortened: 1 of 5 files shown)


kritisingh1:
Sure. I will update the PR with the required changes asap.

test/auth_test.js
```diff
@@ -108,6 +111,7 @@ suite('api/auth', function() {
   testEndpoint({
     method: 'get',
     route:  '/test-deprecated-satisfies',
+    name: placeholderArray[1],
```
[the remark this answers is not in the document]
kritisingh1: Yeah. Its the same way. I  had initially thought if I could pass the values of the new array in a loop. But then found thats not possible. :sweat_smile: 
Nonetheless, I'm updating this.

kritisingh1:
@djmitche Updated. Please review :)

src/api.js
```diff
@@ -789,7 +789,13 @@ API.prototype.declare = function(options, handler) {
   if (this._entries.filter(entry => entry.route == options.route && entry.method == options.method).length > 0) {
     throw new Error('Identical route and method declaration.');
   }
-  this._entries.push(options);
+  if (this._entries.map(a => a.name).some(function(element) {
+    return element == options.name;
+  })) {
```
djmitche: In ES6, you can use arrow functions instead of `function`, and they fit on a single line a lot better.  Also, there's no need to do the mapping separately from the check.  How about
`if (this._entries.some(entry => entry.name === options.name))`

src/api.js
```diff
@@ -789,7 +789,13 @@ API.prototype.declare = function(options, handler) {
...
+    throw new Error('This function has already been declared.');
+  } else {
+    this._entries.push(options);
+  }
```
djmitche: The true branch of this conditional doesn't return (it throws), so there's no need for an `else` here.

src/api.js
```diff
@@ -789,6 +789,9 @@ API.prototype.declare = function(options, handler) {
   if (this._entries.filter(entry => entry.route == options.route && entry.method == options.method).length > 0) {
     throw new Error('Identical route and method declaration.');
   }
+  if (this._entries.some(entry => entry.name === options.name)) {
+    throw new Error('This function has already been declared.');
+  }
```
djmitche: :1st_place_medal:

PR #98, opened 2018-05-17
Redeployability Great Library Update - 10.0

djmitche:
This may not pass its tests, since it requires newer versions of other libraries to be landed, and those requirements are circular.

On a positive review, this will be squashed into a single commit, landed, all other GLU10 libraries `yarn upgrade`'d to 10.0.0, and this released as 10.0.0.

Background is in https://public.etherpad-mozilla.org/p/library-updates


src/service.js
```diff
@@ -0,0 +1,879 @@
...
+        `Context has unexpected property: ${property}`);
+    });
+
+    // make `entries` specific to this rootUrl
+    this.entries = options.api.entries.map((entry) => {
+      entry = _.clone(entry);
+
+      // fully-qualify schema references
+      if (entry.input) {
+        assert(!entry.input.startsWith('http'), 'entry.input should be a filename, not a url');
```
imbstack: Were we doing this in any places before?

imbstack: Oh, I see this down below now. Really interesting that we ever did things like that.

PR #100, opened 2018-05-22
Bug 1463260 - Fix rootUrl for docs writing

imbstack:
Tests will fail until we land taskcluster/taskcluster-lib-validate#39


src/api.js
```diff
@@ -396,7 +298,7 @@ var schema = function(validate, options) {
         return res.reportError(
           'InputValidationError',
           error,
-          {schema: options.input});
+          {schema: `${rootUrl}${options.input}`});
```
djmitche: Sorry, forgot to make this comment - could we use [`url.resolve`](https://nodejs.org/api/url.html#url_url_resolve_from_to) here instead, to make it clear these are relative URIs?

PR #109, opened 2018-07-14
Fix errors in error handling

djmitche:
Stop assigning to a const (!), and also report the bad message.

@ydidwania


djmitche:
v12.3.1

PR #112, opened 2018-08-24
Bug 1484702 - don't include trailing / in baseUrl

djmitche:
There are no actual tests for this, but I verified it manually.

### DIFF
--- a/src/builder.js
+++ b/src/builder.js
@@ -185,7 +185,7 @@ class APIBuilder {
       title:              this.title,
       description:        this.description,
       // We hardcode taskcluster.net here because no other system uses baseUrl
-      baseUrl:            libUrls.api('https://taskcluster.net', this.serviceName, this.version, ''),
+      baseUrl:            `https://${this.serviceName}.taskcluster.net/${this.version}`,
       serviceName:        this.serviceName,
       entries: this.entries.filter(entry => !entry.noPublish).map(entry => {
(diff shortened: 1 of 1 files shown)


src/builder.js
```diff
@@ -185,7 +185,7 @@ class APIBuilder {
       title:              this.title,
       description:        this.description,
       // We hardcode taskcluster.net here because no other system uses baseUrl
-      baseUrl:            libUrls.api('https://taskcluster.net', this.serviceName, this.version, ''),
+      baseUrl:            `https://${this.serviceName}.taskcluster.net/${this.version}`,
```
petemoore: What is the reason not to use libUrls?

petemoore: I think I'd prefer to implement `baseUrl` as a method in libUrls library, or continue to use `libUrls.api` but change that method to only include a `/` if `path` is a non-empty string.

It feels like otherwise we could duplicate this logic in other places, when really the logic for determining the baseUrl should be a function of the taskcluster-lib-urls library.

What do you think?

djmitche: `baseUrl` is a deprecated concept, and libUrls implements the functionality that replaces it (rootUrl).  The existing support for baseUrl is going away as soon as the clients don't need it anymore.  So, this is a quick fix to a bug in the temporary backward-compatibility code for taskcluster.net.  I think adding code to libUrls to fix this bug, only to then remove it in a few weeks, would be putting that fix in the wrong place.

PR #113, opened 2018-08-28
display client id for insufficient scopes

kritisingh1:
Reference : Bug 1457126
@djmitche Please review

### DIFF
--- a/test/auth_test.js
+++ b/test/auth_test.js
@@ -316,9 +316,7 @@ suite('api/auth', function() {
     scopes: {AllOf: [{for: 'scope', in: 'scopes', each: '<scope>'}]},
     handler: async (req, res) => {
       res.status(200).json({
-        /*scopes: await req.scopes(),*/
         clientId: await req.clientId(),
-        /*expires: await req.expires(),*/
       });
     },
     tests : [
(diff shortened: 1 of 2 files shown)


djmitche:
Sorry, somehow I missed this.  I'll look today.

test/auth_test.js
```diff
@@ -309,6 +309,40 @@ suite('api/auth', function() {
...
+  testEndpoint({
+    method: 'get',
+    route: '/test-id-scope',
+    name: 'test-id-scope',
+    scopes: {AllOf: [{for: 'scope', in: 'scopes', each: '<scope>'}]},
+    handler: async (req, res) => {
+      res.status(200).json({
+        /*scopes: await req.scopes(),*/
+        clientId: await req.clientId(),
+        /*expires: await req.expires(),*/
```
djmitche: please remove the commented-out code

test/auth_test.js
```diff
@@ -309,6 +309,40 @@ suite('api/auth', function() {
...
+            return res;
+          }),
+      },
+      {
+        label: 'client does not have sufficient scopes',
+        id: 'test-user',
+        tester: (auth, url) => request.get(url).hawk(auth)
+          .then(function(res) {
+            assert(res.body.clientId != 'admin');
+            return res;
```
djmitche: I think this should also assert that there is an error message containing "Client ID test-user does not "...

kritisingh1:
@djmitche sorry for the silence. I've been stuck at this for quite a while. Not sure I have done this correctly yet. Please review. Thanks.

test/auth_test.js
```diff
@@ -309,6 +309,58 @@ suite('api/auth', function() {
...
+    scopes: {AllOf: ['service:<param>']},
+    handler: async (req, res) => {
+      try {
+        await req.authorize({param: 'myfolder/other-resource'});
+        res.reply({});
+      } catch (err) {
+        if (err.code === 'AuthorizationError') {
+          assert(err.messageTemplate.indexOf('Client ID nobody does not have sufficient scopes')!==-1);
+          return res.reportError('InsufficientScopes', err.messageTemplate, err.details);
+        }
```
djmitche: This tests the right thing, in that without your change to `auth.js` the test will fail.  But, it runs in the "server" (not really a server since it's a test and everything's on the same computer, but pretend..), rather than on the client.  Still, checking it on the "client" would require making bigger changes to how `testEndpoint` is implemented, which would be a much bigger change.  So this is good :)

PR #114, opened 2018-08-31
Bug-1347158 : Added http response json schema for ping endpoint

Saurabh-Pakhare:
Bug-1347158 (https://bugzilla.mozilla.org/show_bug.cgi?id=1347158#c1)
Added json schema for the pint endpoint.

### DIFF
--- a/src/schemas/ping-response.json
+++ b/src/schemas/ping-response.json
@@ -1,0 +1,17 @@
+{
+  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "title" : "Ping Response",
+  "description" : "Response to a Ping request which tells service status and uptime",
+  "type" : "object",
+  "properties" : {
+    "alive" : {
+      "description": "Denotes that the server is up and running",
+      "type" : "boolean"
+    },
+    "uptime" : {
+      "description": "Uptime of the server",
+      "type" : "number"
+    }
+  },
+  "required" : ["alive", "uptime"]
+}


ccooper:
@Saurabh-Pakhare Are you planning to continue with this?

Saurabh-Pakhare:
@ccooper Yes I will be working on it now, apologies for the delay

Saurabh-Pakhare:
@djmitche So it will not load schemas for any other api as well?

djmitche:
Right, as-implemented right now it does not.

Here's what I would suggest: in taskcluster-lib-validate, add a method to SchemaSet that will load additional schemas from a given directory and return a *new* SchemaSet object containing both the original and newly-loaded schemas.  Then we can use that method in taskcluster-lib-api.

Saurabh-Pakhare:
I think that adding such method in taskcluster-lib-validate SchemaSet would need some refactoring, can't we simply move the schemas directory at the project root level, wouldn't that solve the problem?

djmitche:
I'm not sure what you're suggesting -- perhaps give it a try and see how it works?

Saurabh-Pakhare:
@djmitche In a previous comment you mentioned this 
> the SchemaSet that is passed to the APIBuilder does not load schemas from this directory
(src/schemas in this library), so just adding output: 'ping.json#' will not do the trick.

And the taskcluster-lib-validate's README also says this 
> Schemas are typically stored under schemas/ in the repository root, with a directory per schema version.

So i was thinking in taskcluster-lib-api can we move the schema from `src/schemas` to taskcluster-lib-api/schemas/{version}, i.e taking it out from src and move it in schemas folder at root level so then taskcluster-lib-validate will be able to load it as it does in other projects

djmitche:
Let's see if it works?

djmitche:
Let's see if it works?

Saurabh-Pakhare:
Yeah so for testing it I will have to integrate with taskcluster-lib-validate locally right?

PR #116, opened 2018-09-04
Pass sourceIp to taskcluster-auth

ajvb:
Pass the original source IP of the request to taskcluster-auth so that it can log it within it's auditlog.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1488545
Related: https://github.com/taskcluster/taskcluster-auth/pull/163

### DIFF
--- a/src/middleware/auth.js
+++ b/src/middleware/auth.js
@@ -164,6 +164,7 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
       host:             host.name,
       port:             parseInt(port, 10),
       authorization:    req.headers.authorization,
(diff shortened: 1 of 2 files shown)


src/middleware/auth.js
```diff
@@ -164,6 +164,7 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
       host:             host.name,
       port:             parseInt(port, 10),
       authorization:    req.headers.authorization,
+      sourceIp:         req.ip,
```
djmitche: we have 'trust proxy' enabled in Express, so this should give us the right data.

PR #122, opened 2018-09-27
add apiVersion to the list of required properties

djmitche:
I noticed this was missing when backporting for aws-provisioner.

I'll upload to S3 and copy to tc-references when merging.


djmitche:
merged, uploaded, and copied.

PR #124, opened 2018-09-28
Allow bots to use tc

imbstack:
Bots don't have valid usernames for emails I guess?

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -19,9 +19,11 @@ tasks:
       metadata:
         name: "taskcluster-lib-api test"
         description: "tests for metrics and monitoring library"
-        owner: ${event.pull_request.user.login}@users.noreply.github.com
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -19,9 +19,11 @@ tasks:
       metadata:
         name: "taskcluster-lib-api test"
         description: "tests for metrics and monitoring library"
-        owner: ${event.pull_request.user.login}@users.noreply.github.com
+        owner:
+          $if: 'event.pull_request.user.type == "User"'
+          then: ${event.pull_request.user.login}@users.noreply.github.com
```
owlishDeveloper: There's no `event.pull_request.user` in the bot payload?

owlishDeveloper:
Cool! (I initially had a question but then I looked closer)

PR #134, opened 2018-10-29
Bug 1502892 - Queue not calling req.authorize

jhford:


jhford:
Unit tests pass, but I haven't really exercised this patch much.  I think it should be safe to land.

imbstack:
I have published this in `v12.5.2`

PR #143, opened 2018-11-20
Bug 1508846 - generate new-format references

djmitche:


djmitche:
It will happen during the build of tc-references.  I have plans to support doing that earlier, by using tc-lib-references as a devDependency and constructing a single-service References instance, then validating that.  That will come later.

PR #147, opened 2018-12-07
res.errorReport internal return

arshadkazmi42:
Fixes [bugzilla#1437191](https://bugzilla.mozilla.org/show_bug.cgi?id=1437191)

### DIFF
--- a/src/middleware/auth.js
+++ b/src/middleware/auth.js
@@ -5,6 +5,7 @@ const crypto = require('crypto');
 const utils = require('../utils');
 const ScopeExpressionTemplate = require('../expressions');
 const Debug = require('debug');
+const ErrorReply = require('../error-reply');
 
 /* In production, log authorizations so they are included in papertrail regardless of
  * DEBUG settings; otherwise, log with debug
@@ -315,11 +316,11 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
       }
     } catch (err) {
       if (err.code === 'AuthorizationError') {
-        return res.replyError('InsufficientScopes', err.messageTemplate, err.details);
+        return next(new ErrorReply({code: 'InsufficientScopes', message: err.messageTemplate, details: err.details}));
       } else if (err.code === 'AuthenticationError') {
-        return res.replyError('AuthenticationFailed', err.message, err.details);
+        return next(new ErrorReply({code: 'AuthenticationFailed', message: err.message, details: err.details}));
       }
-      return res.replyError(err);
+      return next(new ErrorReply(err));
     };
   };
 };
--- a/src/middleware/auth.js
+++ b/src/middleware/auth.js
@@ -315,11 +315,11 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
       }
     } catch (err) {
       if (err.code === 'AuthorizationError') {
-        return res.reportError('InsufficientScopes', err.messageTemplate, err.details);
+        return res.replyError('InsufficientScopes', err.messageTemplate, err.details);
       } else if (err.code === 'AuthenticationError') {
-        return res.reportError('AuthenticationFailed', err.message, err.details);
+        return res.replyError('AuthenticationFailed', err.message, err.details);
       }
-      return res.reportInternalError(err);
+      return res.replyError(err);
     };
   };
 };
--- a/src/middleware/auth.js
+++ b/src/middleware/auth.js
@@ -320,7 +320,7 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
       } else if (err.code === 'AuthenticationError') {
         return next(new ErrorReply({code: 'AuthenticationFailed', message: err.message, details: err.details}));
       }
-      return next(new ErrorReply(err));
+      return next(err);
     };
   };
 };
--- a/src/middleware/errors.js
+++ b/src/middleware/errors.js
@@ -1,6 +1,3 @@
-const uuid = require('uuid');
-const debug = require('debug')('api:errors');
-const _ = require('lodash');
 const ErrorReply = require('../error-reply')
 
 const ERROR_CODES = {
@@ -28,87 +25,14 @@ exports.ERROR_CODES = ERROR_CODES;
  * allowed error codes to HTTP status codes, and `monitor` is an instance of
  * `raven.Client` from an instance of taskcluster-lib-monitor.
  */
-const buildReportErrorMethod = ({errorCodes, monitor, entry}) => {
-  const {name: method, cleanPayload} = entry;
+const buildReportErrorMethod = () => {
   return (req, res, next) => {
     res.reportError = (code, message, details = {}) => {
-      let status = errorCodes[code];
-      let payload = req.body;
-      if (cleanPayload) {
-        payload = cleanPayload(payload);
-      }
-
-      if (status === undefined || typeof message !== 'string') {
-        const newMessage = 'Internal error, unknown error code: ' + code + '\n' +
-          (message || 'Missing message!');
-        code = 'InternalServerError';
-        status = 500;
-        if (monitor) {
-          const err = new Error(newMessage);
-          err.badMessage = message;
-          err.badCode = code;
-          err.details = details;
-          monitor.reportError(err);
-        }
-        message = newMessage;
-      }
-
-      const requestInfo = {
-        method,
-        params:  req.params,
-        payload,
-        time:    (new Date()).toJSON(),
-      };
-
-      message = message.replace(/{{([a-zA-Z0-9_-]+)}}/g, (text, key) => {
-        let value = details.hasOwnProperty(key) ? details[key] : text;
-        if (typeof value !== 'string') {
-          value = JSON.stringify(value, null, 2);
-        }
-        return value;
-      }) + [
-          '\n\n---\n',
-          '* method:     ' + method,
-          '* errorCode:  ' + code,
-          '* statusCode: ' + status,
-          '* time:       ' + requestInfo.time,
-        ].join('\n');
-        throw ({next, code, message, requestInfo});
+      throw new ErrorReply({code, message, details});
     };
 
-    res.reportInternalError = (err, tags = {}) => {
-      try {
-        const incidentId = uuid.v4();
-        res.reportError(
-          'InternalServerError',
-          'Internal Server Error, incidentId: ' + incidentId,
-          {incidentId}
-        );
-        debug(
-          'Error occurred handling: %s, err: %s, as JSON: %j, incidentId: %s',
-          req.url, err, err, incidentId, err.stack
-        );
-        if (monitor) {
-          err.incidentId = incidentId;
-          err.method = method;
-          err.params = req.params;
-          let payload = req.body;
-          if (cleanPayload) {
-            payload = cleanPayload(payload);
-          }
-          err.payload = req.payload;
-          monitor.reportError(err, Object.assign(tags, {method}));
-        }
-      } catch (err) {
-        return res.replyError(err);
-      }
-    };
-
-    res.replyError = ({code, message, requestInfo}) => {
-      if (ERROR_CODES[code]) {
-        return res.status(ERROR_CODES[code]).json({code, message, requestInfo})
-      }
-      return res.reportInternalError({code, message, requestInfo})
+    res.replyError = (code, message, details) => {
+      return next(new ErrorReply({code, message, details}));
     }
     next();
   };
--- a/src/middleware/errors.js
+++ b/src/middleware/errors.js
@@ -72,30 +72,36 @@ const buildReportErrorMethod = ({errorCodes, monitor, entry}) => {
           '* statusCode: ' + status,
           '* time:       ' + requestInfo.time,
         ].join('\n');
-      throw ({status, code, message, requestInfo});
+        throw ({status, code, message, requestInfo});
     };
 
     res.reportInternalError = (err, tags = {}) => {
-      const incidentId = uuid.v4();
-      res.reportError(
-        'InternalServerError',
-        'Internal Server Error, incidentId: ' + incidentId,
-        {incidentId}
-      );
-      debug(
-        'Error occurred handling: %s, err: %s, as JSON: %j, incidentId: %s',
-        req.url, err, err, incidentId, err.stack
-      );
-      if (monitor) {
-        err.incidentId = incidentId;
-        err.method = method;
-        err.params = req.params;
-        let payload = req.body;
-        if (cleanPayload) {
-          payload = cleanPayload(payload);
+      try {
+        const incidentId = uuid.v4();
+        res.reportError(
+          'InternalServerError',
+          'Internal Server Error, incidentId: ' + incidentId,
+          {incidentId}
+        );
+        debug(
+          'Error occurred handling: %s, err: %s, as JSON: %j, incidentId: %s',
+          req.url, err, err, incidentId, err.stack
+        );
+        if (monitor) {
+          err.incidentId = incidentId;
+          err.method = method;
+          err.params = req.params;
+          let payload = req.body;
+          if (cleanPayload) {
+            payload = cleanPayload(payload);
+          }
+          err.payload = req.payload;
+          monitor.reportError(err, Object.assign(tags, {method}));
         }
-        err.payload = req.payload;
-        monitor.reportError(err, Object.assign(tags, {method}));
+      } catch (err) {
+        const status = err.status;
+        delete err.status
+        return res.status(status).json(err);
       }
     };
     next();
--- a/src/middleware/errors.js
+++ b/src/middleware/errors.js
@@ -1,6 +1,7 @@
 const uuid = require('uuid');
 const debug = require('debug')('api:errors');
 const _ = require('lodash');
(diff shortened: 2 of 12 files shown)


arshadkazmi42:
@djmitche Added my code here, for review.
At present many other tests are failing.

src/middleware/errors.js
```diff
@@ -72,7 +72,7 @@ const buildReportErrorMethod = ({errorCodes, monitor, entry}) => {
           '* statusCode: ' + status,
           '* time:       ' + requestInfo.time,
         ].join('\n');
-      res.status(status).json({code, message, requestInfo});
```
djmitche: This part looks good!

src/middleware/handle.js
```diff
@@ -31,7 +31,7 @@ const callHandler = ({entry, context, monitor}) => {
       } else if (err.code === 'AuthenticationError') {
         return res.reportError('AuthenticationFailed', err.message, err.details);
       }
-      return res.reportInternalError(err);
```
djmitche: I suspect this is the issue behind the test failures.  You can probably find more by running a single test with `DEBUG=*`.  For example:

```shell
DEBUG=* mocha test/auth_test.js --grep 'function that still uses satisfies fails'
```

arshadkazmi42: I am kinda stuck at this. Tried running alot of tests. I tried adding this here
```
if (err.status) {
        return res.status(err.status).json(err);
      }
      return res.reportInternalError(err);
```
but that also doesn't seems to work. 
Can you give me some info on where exactly should I look for?

djmitche: I think this is a case for tracing through what happens by hand.  Let's take, for example, code in an API handler that calls
```js
res.reportError(
  'TooManyFoos',
  'You can only have 3 foos.  These foos already exist:\n```\n{{foos}}\n```',
  {foos: foomanager.foos(request.fooId)});
```

That will throw (from `errors.js`):
```js
       throw ({status, code, message, requestInfo}); 
```

That will be caught (in `handle.js`), and since `err.code` is not `AuthorizationError` or `AuthenticationError`, will execute:
```js
return res.status(err.status).json(err);
```
which is close to correct, although it contains an extra `status` property in the response body that should not be present.

Here are a few more scenarios to trace through:
 * some other exception occurs in the API handler (for example, `throw new Error('oh no!')`)
 * something calls `res.reportError('AuthorizationError', ..)` (this code is handled specially in `handle.js`), or similarly `AuthenticationError`

Once you've done that, what needs to be done should be a little clearer.

arshadkazmi42: I did some debugging, got 1 doubt. 
For this test
`https://github.com/taskcluster/taskcluster-lib-api/blob/master/test/errors_test.js#L142`
in this test, throw in `errors.js` is not going to `handle.js`.
Is all the request does not goes to `handle.js` due to which correct request is not returned and this test has failed.

arshadkazmi42: This makes sense.

I found one more test which is giving similar issue.
https://github.com/taskcluster/taskcluster-lib-api/blob/master/test/errors_test.js#L67
The test at above line,
it throws the error ` Error('oh no!')` so for building the 500 response it needs to go to `reportInternalError` which is then calling `res.reportError` and when it throws from `reportError` it crashes as `reportInternalError` does not have a catch block.
So I added a try catch block wrapping `reportInternalError` code.
Doing the above change, for separating "reporting an error" and "replying with an error" can be used for this case also, that is, we can call "reporting an error" from `reportInternalError` catch block.

Any thoughts on this? Does this approach makes sense? Or any inputs from your end?

djmitche: Let's see some code :)

djmitche:
Keep working at it :)

arshadkazmi42:
@djmitche I have updated the code, all tests have passed. Can you have a look at the changes, if there is any modification needed.

djmitche:
will do!

arshadkazmi42:
@djmitche So I created an error handler and an error reply class. 
But problem is this
https://github.com/taskcluster/taskcluster-lib-api/pull/147/files#diff-d8684581c724b6e8d754a2bcce3c7c3fR76
Due to this throw, any middleware which is calling it needs a try catch to handle it.
When I change this to  
`next({code, message, requestInfo})`
it works fine but it does not fixes the issue, which is mentioned here
https://bugzilla.mozilla.org/show_bug.cgi?id=1437191#c6

src/middleware/errors.js
```diff
@@ -72,32 +73,43 @@ const buildReportErrorMethod = ({errorCodes, monitor, entry}) => {
...
     };
+
+    res.replyError = ({code, message, requestInfo}) => {
+      if (ERROR_CODES[code]) {
+        return res.status(ERROR_CODES[code]).json({code, message, requestInfo})
+      }
+      return res.reportInternalError({code, message, requestInfo})
+    }
     next();
   };
```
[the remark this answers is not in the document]
arshadkazmi42: Did you mean `next()` should we remove it?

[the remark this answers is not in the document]
djmitche: I mean the whole `res.replyError` method.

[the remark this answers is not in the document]
arshadkazmi42: I am using that function in my current fix, to trigger the express-error handler using `next(err)`.
Can you review my latest changes

src/middleware/errors.js
```diff
@@ -27,77 +25,15 @@ exports.ERROR_CODES = ERROR_CODES;
...
-        let payload = req.body;
-        if (cleanPayload) {
-          payload = cleanPayload(payload);
-        }
-        err.payload = req.payload;
-        monitor.reportError(err, Object.assign(tags, {method}));
-      }
-    };
+    res.replyError = (code, message, details) => {
+      return next(new ErrorReply({code, message, details}));
```
djmitche: This `next` method is for the `buildReportErrorMethod` middleware, not for the middleware that called `replyError`.  That might work now, but it's probably not something Express intends to support.

src/middleware/auth.js
```diff
@@ -315,11 +315,11 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
       }
     } catch (err) {
       if (err.code === 'AuthorizationError') {
-        return res.reportError('InsufficientScopes', err.messageTemplate, err.details);
+        return res.replyError('InsufficientScopes', err.messageTemplate, err.details);
       } else if (err.code === 'AuthenticationError') {
-        return res.reportError('AuthenticationFailed', err.message, err.details);
+        return res.replyError('AuthenticationFailed', err.message, err.details);
       }
-      return res.reportInternalError(err);
```
djmitche: I think all three of these can be turned into `next(err)` or `next(new ErrorReply(..))`.

src/middleware/handle.js
```diff
@@ -27,11 +27,11 @@ const callHandler = ({entry, context, monitor}) => {
       }
     }).catch((err) => {
       if (err.code === 'AuthorizationError') {
-        return res.reportError('InsufficientScopes', err.messageTemplate, err.details);
+        return res.replyError('InsufficientScopes', err.messageTemplate, err.details);
       } else if (err.code === 'AuthenticationError') {
-        return res.reportError('AuthenticationFailed', err.message, err.details);
+        return res.replyError('AuthenticationFailed', err.message, err.details);
       }
-      return res.reportInternalError(err);
```
djmitche: All three of these, too, can be turned into calls to `next`.

src/middleware/express-error.js
```diff
@@ -0,0 +1,70 @@
...
+ * parameter to regular expression or function that returns a message as string
+ * if the parameter is invalid.
+ *
+ * Parameters not listed in `req.params` will be ignored. But parameters
+ * present must match the pattern given in `options` or the request will be
+ * rejected with a 400 error message.
+ */
+const expressError = ({errorCodes, entry, monitor}) => {
+  const {name: method, cleanPayload} = entry;
+  return (err, req, res, next) => {
```
djmitche: I don't see a special-case for the ErrorReply type here (`err instanceof ErrorReply`).  Instead, it depends on the fact that such objects have `code`, `details`, and `message` -- but lots of errors have `code` and `message`, and it's not unreasonable to think one might have `details`.  This would result in showing that data to the caller, possibly revealing secret information in `details`.

A better separation would be something like this:

```js
if (!err instanceof ErrorReply) {
  // convert an arbitrary error into an ErrorReply, obscuring any details of the error
  // as they might reveal useful information to an attacker
  // .. call monitor.reportError like the old reportInternalError did
  // .. create an incidentId and set `err` to a new ErrorReply similar to how reportInternalError called reportError
}
// .. continue on as below, now knowing that `err` is an ErrorReply object
```

src/middleware/auth.js
```diff
@@ -315,11 +316,11 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
...
     } catch (err) {
       if (err.code === 'AuthorizationError') {
-        return res.replyError('InsufficientScopes', err.messageTemplate, err.details);
+        return next(new ErrorReply({code: 'InsufficientScopes', message: err.messageTemplate, details: err.details}));
       } else if (err.code === 'AuthenticationError') {
-        return res.replyError('AuthenticationFailed', err.message, err.details);
+        return next(new ErrorReply({code: 'AuthenticationFailed', message: err.message, details: err.details}));
       }
-      return res.replyError(err);
+      return next(new ErrorReply(err));
```
djmitche: This one is *not* an ErrorReply -- we already have an error that we need to handle.  So just `next(err)`.

src/middleware/handle.js
```diff
@@ -27,11 +28,11 @@ const callHandler = ({entry, context, monitor}) => {
...
     }).catch((err) => {
       if (err.code === 'AuthorizationError') {
-        return res.replyError('InsufficientScopes', err.messageTemplate, err.details);
+        return next(new ErrorReply({code: 'InsufficientScopes', message: err.message, details: err.details}));
       } else if (err.code === 'AuthenticationError') {
-        return res.replyError('AuthenticationFailed', err.message, err.details);
+        return next(new ErrorReply({code: 'AuthenticationFailed', message: err.message, details: err.details}));
       }
-      return res.replyError(err.code, err.message, err.details)
+      return next(new ErrorReply(err));
```
djmitche: same here..

src/middleware/express-error.js
```diff
@@ -14,6 +16,29 @@ const expressError = ({errorCodes, entry, monitor}) => {
   return (err, req, res, next) => {
 
     const incidentId = uuid.v4();
+    if (!(err instanceof ErrorReply)) {
+      res.reportError(
+        'InternalServerError',
+        'Internal Server Error, incidentId: ' + incidentId,
+        {incidentId}
+      );
```
djmitche: Calling reportError will just throw another error, which is not what we want.  Instead, we need to replace `err` here with an ErrorReply containing `InternalServerError` and an incidentId and so on.

src/middleware/express-error.js
```diff
@@ -14,6 +16,29 @@ const expressError = ({errorCodes, entry, monitor}) => {
...
+      if (monitor) {
+        err.incidentId = incidentId;
+        err.method = method;
+        err.params = req.params;
+        let payload = req.body;
+        if (cleanPayload) {
+          payload = cleanPayload(payload);
+        }
+        err.payload = req.payload;
+        monitor.reportError(err, Object.assign(tags, {method}));
```
djmitche: `tags` isn't defined here.  The second argument to `monitor.reportError` gives tags to assign to the error report, and in this case the only tag we want to add is `{method}`, so you can omit the `Object.assign` bit.

src/middleware/express-error.js
arshadkazmi42: `reportError` is converting this into an ErrorReply. if we are passing ErrorReply from here, we might have to remove ErrorReply from `reportError`

djmitche: I'm not sure what you mean.  It is converting it into an ErrorReply, but then throwing it.

arshadkazmi42: reportError is doing something like this

```
res.reportError = (code, message, details = {}) => {
      throw new ErrorReply({code, message, details});
    };
```

arshadkazmi42: I read your comment and again, I think I got it wrong previously. 
Just to be clear, you meant to change this to something like this 
```
err = new ErrorReply({
        code: 'InternalServerError',
        message: 'Internal Server Error, incidentId: ' + incidentId,
        details: {incidentId}
      });
```
Correct??

src/middleware/express-error.js
```diff
@@ -0,0 +1,95 @@
...
+        if (cleanPayload) {
+          payload = cleanPayload(payload);
+        }
+        err.payload = req.payload;
+        monitor.reportError(err, {method});
+      }
+    }
+
+    let code = err.code || 'InternalServerError';
+    let details = err.details || {incidentId};
```
djmitche: There's no need for these `|| ..` anymore -- `err` is always an ErrorReply, and ErrorReply will always have these properties.

src/middleware/express-error.js
```diff
@@ -0,0 +1,95 @@
...
+    let message = err.message;
+    if (!err.code) {
+      message = 'Internal Server Error, incidentId: ' + incidentId;
+    }
```
djmitche: ..and similarly, this can be omitted (since it will always have a code)

src/middleware/express-error.js
```diff
@@ -0,0 +1,92 @@
...
+      if (monitor) {
+        err.incidentId = incidentId;
+        err.method = method;
+        err.params = req.params;
+        let payload = req.body;
+        if (cleanPayload) {
+          payload = cleanPayload(payload);
+        }
+        err.payload = req.payload;
+        monitor.reportError(err, {method});
```
djmitche: So I was about to mark this "approved", but I thought I'd just try it out testing a real application.  Happiliy, all of the tests passed

However, I added a `throw new Error('uhoh!');` to an API method in taskcluster-secrets -- just to test handling of internal errors -- and here's what was logged:
```
  api:errors Error occurred handling: /secret/captain%3AXkyBT0SfTW6kBYiyfwGLBw, err: Error: Internal Server Error, incidentId: 4cddbbce-d531-414e-89ce-8dfe8333ad60, as JSON: {"code":"InternalServerError","message":"Internal Server Error, incidentId: 4cddbbce-d531-414e-89ce-8dfe8333ad60","details":{"incidentId":"4cddbbce-d531-414e-89ce-8dfe8333ad60"}}, incidentId: 4cddbbce-d531-414e-89ce-8dfe8333ad60 Error: Internal Server Error, incidentId: 4cddbbce-d531-414e-89ce-8dfe8333ad60
    at /home/dustin/p/taskcluster-lib-api/src/middleware/express-error.js:20:13
    at Layer.handle_error (/home/dustin/p/taskcluster-lib-api/node_modules/express/lib/router/layer.js:71:5)
    at next (/home/dustin/p/taskcluster-lib-api/node_modules/express/lib/router/route.js:135:13)
    at Promise.resolve.then.then.catch (/home/dustin/p/taskcluster-lib-api/src/middleware/handle.js:35:14)
    at <anonymous>
    at runMicrotasksCallback (internal/process/next_tick.js:122:5)
    at _combinedTickCallback (internal/process/next_tick.js:132:7)
    at process._tickDomainCallback (internal/process/next_tick.js:219:9) +816ms
  taskcluster-lib-monitor reportError: Error: Internal Server Error, incidentId: 4cddbbce-d531-414e-89ce-8dfe8333ad60 level=[object Object] tags={} Error: Internal Server Error, incidentId: 4cddbbce-d531-414e-89ce-8dfe8333ad60
    at /home/dustin/p/taskcluster-lib-api/src/middleware/express-error.js:20:13
    at Layer.handle_error (/home/dustin/p/taskcluster-lib-api/node_modules/express/lib/router/layer.js:71:5)
    at next (/home/dustin/p/taskcluster-lib-api/node_modules/express/lib/router/route.js:135:13)
    at Promise.resolve.then.then.catch (/home/dustin/p/taskcluster-lib-api/src/middleware/handle.js:35:14)
    at <anonymous>
    at runMicrotasksCallback (internal/process/next_tick.js:122:5)
    at _combinedTickCallback (internal/process/next_tick.js:132:7)
    at process._tickDomainCallback (internal/process/next_tick.js:219:9) +814ms
```

So it's logging the "Internal Server Error" both with `debug`, the first section, and with `monitor.reportError`, the second section.  It's much more useful for debugging to log the *original* error, and only send the "Internal Server Error" to the API client.  Can you rearrange a little to make that work?

djmitche: Thanks for moving the `debug` call up.  I added a commit to also move the `monitor.reportError` call up (and add some comments).

arshadkazmi42:
:tada: :tada:

djmitche:
Indeed, nice work on this -- thanks for your patience through multiple rounds of review!

arshadkazmi42:
I enjoyed working on this. Thank you for your valuable reviews. :+1:

djmitche:
@arshadkazmi42 so I just noticed there are still some instances of reportInternalError in the source!

djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1515841

arshadkazmi42:
I look into it asap. Thank you for finding it out.

PR #149, opened 2018-12-21
Bug 1515841 - remove remaining uses of reportInternalError

djmitche:


djmitche:
(please land and release v12.8.1 if this is OK)

djmitche:
I should add tests for this..

djmitche:
@imbstack note that this is badly broken in the latest version of tc-lib-api, so a bit urgent.

djmitche:
v12.8.1.  Won't it be nice when we don't have to mess about with versions?

PR #150, opened 2018-12-21
Updated all reportInternalError reference with next(err)

arshadkazmi42:
Fixes [bugzilla#1515841](https://bugzilla.mozilla.org/show_bug.cgi?id=1515841)

### DIFF
--- a/src/middleware/schema.js
+++ b/src/middleware/schema.js
@@ -66,7 +66,7 @@ const validateSchemas = ({validator, absoluteSchemas, rootUrl, serviceName, entr
     res.reply = (json) => {
       if (!req.hasAuthed) {
         const err = new Error('Deferred auth was never checked!');
-        return next(err);
+        throw err;
       }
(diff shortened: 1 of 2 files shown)


src/middleware/schema.js
```diff
@@ -66,7 +66,7 @@ const validateSchemas = ({validator, absoluteSchemas, rootUrl, serviceName, entr
     res.reply = (json) => {
       if (!req.hasAuthed) {
         const err = new Error('Deferred auth was never checked!');
-        return res.reportInternalError(err);
+        return next(err);
```
djmitche: I think this is another case of calling `next` for one handler when another handler is in progress -- `res.reply` gets called later, after the handler defined on line 44 is complete.  In #149 I used `throw err` here.  That thrown exception will be caught by `handle.js` and passed on to the error handler.

djmitche:
Closed in deference to #150.  Sorry about that, @arshadkazmi42 !

PR #151, opened 2018-12-25
Bug 1514042 - include error message in ISE response in non-production

djmitche:
Contributors often get hung up when modifying an API method, since the API tests go through HTTP and thus all they see for any error is "Internal Server Error".

This includes the error in the HTTP response, and I have verified in a real service that this appears in the output.  The client does automatically retry the 500 a few times, so the tests "hang" briefly, but I think that's survivable.

### DIFF
--- a/test/errors_test.js
+++ b/test/errors_test.js
@@ -1,9 +1,10 @@
-const request         = require('superagent');
-const assert          = require('assert');
-const APIBuilder      = require('../');
-const helper          = require('./helper');
-const _               = require('lodash');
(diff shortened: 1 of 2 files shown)


test/errors_test.js
```diff
@@ -113,6 +118,7 @@ suite('api/errors', function() {
     return request.get(url).then(res => assert(false, 'should have failed!')).catch(res => {
       assert(res.status === 500);
       const response = JSON.parse(res.response.text);
+      console.log(response);
```
imbstack: Left over debug?

